PR #4, opened 2026-03-21
[SURVEY_PROGRAM] Add audiogroup*.dat support

Fancy2209:
Allows audio in SURVEY_PROGRAM to work, (though there seem to be some audio issues as in the Goner Maker you can tell it's Asgore Talking Voice, meaning some effects don't work properly)
It can be seen in action here, before you'd just hear the rude buster sound effect

After  
  
https://github.com/user-attachments/assets/293d1af8-8455-4bfd-b4c5-194ff195f560  

Before  
  
https://github.com/user-attachments/assets/36103f0f-1266-4b90-b54f-fb4281385ed2

### DIFF
--- a/src/audio_system.h
+++ b/src/audio_system.h
@@ -38,5 +38,5 @@ typedef struct {
 
 struct AudioSystem {
     AudioSystemVtable* vtable;
-    DataWin* dataWin;
+    DataWin** audioGroups;
 };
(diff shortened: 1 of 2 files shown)


MrPowerGamerBR:
In the future we'll need to update the PS2 preprocessor to support audio groups, but for now, this is already good :)

PR #5, opened 2026-03-22
feat(vm): add jit support

leap0x7b:
Adds basic JIT support to the VM using sljit, which should theoretically improves performance on some platforms. Works fine on Linux/GLFW and haven't encountered any noticeable problems so far. Couldn't test on PS2 since I can't figure out how to make an ISO for it though since sljit has support for MIPS, it did compile without any problems.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "vendor/sljit"]
+	path = vendor/sljit
+	url = https://github.com/zherczeg/sljit
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,14 @@ file(GLOB PLATFORM_SOURCES src/${PLATFORM}/*.c)
 add_executable(butterscotch ${SOURCES} ${PLATFORM_SOURCES})
 target_include_directories(butterscotch PRIVATE ${CMAKE_SOURCE_DIR}/src)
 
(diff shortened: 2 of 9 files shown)


leap0x7b:
The JIT seems to work on PCSX2. I couldn't really confirm since it's running on an emulator on my slow and aging laptop though Undertale did perform slightly better with JIT and runs at slightly more consistent 30fps. I did encounter some crashes during the Asriel battle though I don't know if it's because of my JIT code or something else and if those crashes also happen on non-JIT as well.

I don't have a real PS2 so it would be really appreciated if someone who has one tested my changes. My neighbor has one though I don't think they would be willing to borrow or give it to me. I have enough money on my wallet to buy a PS2 on an online marketplace though I was saving for a new laptop so I'm not sure if I should buy it or not (+ my parents are very strict so I don't even know if they will even approve of my purchases)

MrPowerGamerBR:
I've tested it on PCSX2 with the latest Butterscotch version + SURVEY_PROGRAM's Jevil fight, but it seems that for some reason it did not provide any meaningful differences (before it was running at 4 FPS, after it was... still running at 4 FPS)

I wonder if something is going wrong for it to not be working

leap0x7b:
will reopen sometime in the future

Issue #24, opened 2026-03-30
Support for MSVC


ske2004:
Hi! I wanna know if MSVC support is out of the question or not. MSVC doesn't support C23 (can be worked around with clang-cl) and doesn't have some POSIX headers (like `getopt.h`)

SergioFLS:
Since 76b5b13 Butterscotch now uses GNU99, which might help on getting it compiling on MSVC.

SergioFLS:
I made my own attempt using real `cl` instead of `clang-cl`: #160

casrielasriel:
done afaik

PR #25, opened 2026-03-31
MSVC example

ske2004:
Leaving here just in case this is just an example

#24 

image

Build command. Require `clang-cl` to be installed, and probably will not work on below VS2026:

```
mkdir build && cd build && cmake .. -DPLATFORM="glfw" -T ClangCL -DCMAKE_MODULE_PATH="C:/Paths/CMakeModules" && msbuild butterscotch.vcxproj
```

Where `C:/Paths/CMakeModules` is path with `Findglfw3.cmake`.


MrPowerGamerBR:
Superseded by https://github.com/ButterscotchRunner/Butterscotch/pull/160

PR #29, opened 2026-04-03
Fix off-by-one sprite frame by advancing image_index at the start of the step.

KSP0O:
**Fix off-by-one sprite frame by advancing image_index at the start of the step.**

The image_index was being advanced after the step events, causing sprites to display one frame ahead on the same step they were updated.

image

### DIFF
--- a/src/runner.c
+++ b/src/runner.c
@@ -1168,6 +1168,27 @@ void Runner_step(Runner* runner) {
         }
     }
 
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
Oh wow, so that's why the frames were wrong, when I was trying to debug this weeks ago I ended stumbling down the path handling code and couldn't figure out what it was

Thanks for your fix!

PR #53, opened 2026-04-21
Replace GLFW with SDL2

Fancy2209:
Untested, but highly requested.

I also fixed the PS2 BC16 build since last commit broke it


coreknee:
itd be nice if we kept both

PR #66, opened 2026-04-28
WIP: Start adding matrix operations

seija-amanojaku:
Marking this as a draft since I still doubt their implementations.


seija-amanojaku:
Those transform operations should work well enough from the quick tests I ran with WWW.

seija-amanojaku:
git is my passion

PR #69, opened 2026-04-28
Fix internal inconsistency in binary read/write endianness

KSP0O:
Fixes an internal inconsistency where some binary paths were explicitly little-endian while others relied on native-endian reads/writes.

### DIFF
--- a/src/binary_reader.c
+++ b/src/binary_reader.c
@@ -1,4 +1,5 @@
 #include "binary_reader.h"
+#include "binary_utils.h"
 #include "utils.h"
 
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


Fancy2209:
I had to fix this for the PS3 port I made but the way you did it is wayy cleaner

MrPowerGamerBR:
One issue is that the PR completely removes the aligned reads from the BinaryUtils (I need to double check, but the reason why it is like that is because the compiler emits less MIPS instructions when it can be sure that it is aligned than when it can't be sure, and because those instructions are in the `executeLoop` hot path, every instruction counts)

PR #84, opened 2026-05-02
ps2/overlay: Remove `fontm` usage

abelbriggs1:
Requires #83.

`fontm` is very convenient for simply printing text, but it is unfortunately not supported by some PS2 hardware, because some PS2 derivatives do not include the OSD. Specific examples include DTL (development kit) consoles and the PSX DVR.

Instead of using `fontm`, we load a custom font from the executable using gsKit.

Unfortunately, gsKit's support for non-`fontm` fonts is very lacking:
- `gsKit_init_font_raw()` isn't even linked into the library distributed with PS2SDK, so we have to manually create the structure.
- `gsKit_font_print_scaled()` doesn't support a bunch of stuff that the `fontm` API does
  - spacing adjustment
  - `\t` and ` ` characters
  - alignment

So the implementation here is a bit scuffed. But it does function, for the most part.

Examples:

image

image

The current font is Arial and was taken from the `gsKit` examples, but it should be easily replacable if desired. Note that the font texture needs to be `GS_PSM_CT32` or gsKit will not upload it correctly.


abelbriggs1:
Rebased on latest main now that #83 is merged.

MrPowerGamerBR:
I've decided to use a pre-created 256x256 JetBrains Mono font atlas and skip using it via gsKit altogether, because I thought "hmm, the PlayStation 3 target also will need a debug font atlas too, so it would be better to store it in a way that the atlas data (like the glyphs) could be reused between the PlayStation 2 and PlayStation 3 targets..."

image

image

But after doing all of that, I've noticed that my changes ended up looking similar to yours, whoops *insert sad cat emoji here*

However, I do think that the "purpose" of it is the same, and I think that now Butterscotch may work on PS2 TOOL devkits (yay!)

Thanks for your work in any way, because I was never going to know that FONTM wasn't going to work on the PS2 devkits, and other PS2 optimizations tidbits

 https://github.com/MrPowerGamerBR/Butterscotch/commit/f6a2d03ea7dc45f91b4d212bcb231bee5fdb2f2d

PR #87, opened 2026-05-04
Implement audio_exists / sound_exists

LukeWarnut:
Implements audio_exists and alias sound_exists (pre-GMS2).

Used by Minit: Fun Racer, but implementation doesn't result in noticeable changes.

Test program:
[game.ios.zip](https://github.com/user-attachments/files/27372365/game.ios.zip)
Correct output is 1 0 1


LukeWarnut:
Was having second thoughts about the accuracy, but I think I was being thrown off by an inaccuracy in another area.

PR #89, opened 2026-05-05
GLFW 2 support + GNU makefile build + Various legacy support changes

### LINKED ISSUE #88: Possibility to run on Windows 98 / ME (with old hardware)
Hello. As far as PlayStation 2 port of Undertale based on Butterscotch exists, do I understand correctly that Undertale with Butterscotch has much lower system requirements than original Undertale with proprietary GameMaker runtime?
I have a Wyse VX0 thin client (1.2 GHz VIA Eden CPU with actual performance like 600 MHz Pentium, 512 MB RAM, S3 UniChrome Pro graphics adapter with max 128 MB VRAM, Windows 98 SE). Is it theoretically possible to get Undertale running here using Butterscotch?

Un1q32:
Closes #88 
Tested with Debian Linux 4.0 (2007) and Windows 98 SE.
Theoretically should work on as low as Windows NT 3.1, though you'll need some kind of OpenGL driver, and that might be impossible on such an old Windows version.
Tested to compile with as low as GCC 3.1, possibly lower.
glfw2 uses the legacy-gl renderer by default (as it's intended for legacy systems), and doesn't support controllers or headless mode.  Also the only way to build glfw2 is with the makefile build.


TimofeyLednev:
I ported GLFW 2 to Windows NT 3.51 today and It also works on NT 3.51!
изображение

Un1q32:
Did you import your own msvcrt.dll or did you get a toolchain working with crtdll?  Also what GPU did you use that supports NT3?

TimofeyLednev:
> Did you import your own msvcrt.dll or did you get a toolchain working with crtdll? Also what GPU did you use that supports NT3?

I used an already configured modern gcc because I was too lazy to compile my own, I ran it on VMware with Mesa 7.5.2

Un1q32:
Mesa as in software rendered?  Did you need to build mesa yourself?

Un1q32:
I got a crtdll build working, which means possible NT 3.1.

Un1q32:
If you can get me your mesa dll that would be great.

TimofeyLednev:
> If you can get me your mesa dll that would be great.

I just wrote to you on Discord about BeOS and gcc 3.4.0 check Discord (my username timofey38)

TimofeyLednev:
> Mesa as in software rendered? Did you need to build mesa yourself?

Yes software render

PR #103, opened 2026-05-10
Relicense Butterscotch under the MPL-2.0

MrPowerGamerBR:
While I'm quite a open source enjoyer, one real GPL blocker is that the GPL & friends (AGPL and LGPL) is that you can't submit a GPL app to the App Store that you aren't the sole copyright owner, so it wouldn't be possible if in the far future someone creates a App Store app that depends on Butterscotch. ([see the VLC app debacle that happened before for a real case of this happening](https://mailman.videolan.org/pipermail/vlc-devel/2010-November/077027.html), due to that, the [VLC iOS app is licensed under the MPL-2.0](https://github.com/videolan/vlc-ios))

The MPL-2.0 is the "best license" that still maximizes ecosystem openness while being permissive enough that it won't be a issue down the road. More strict than Apache-2.0, which is the "hard floor" because Butterscotch does HEAVILY use GameMaker-HTML5 as a reference, while being below the GPL "soft ceiling".

The tl;dr is that the MPL-2.0 only requires changes that you made to the original files to be open sourced. This does mean that ports wouldn't need to make the platform-specific files open source, which is a bit sad... but if someone does that we can just make our OWN PORT and make it even BETTER. *insert cat sticker here*

The only "dependency" that is GPL-3.0 that we have is UndertaleModTool, but we don't directly depend on it (used only as a reference for the data win parser), so I think it should be A-OK.

For the relicense, everyone here must reply to this PR with "I agree":

@Fancy2209 ✅
@LukeWarnut ✅
@Eliandro4 ✅
@mattx433 ✅
@seija-amanojaku ✅
@KSP0O ✅
@Classic0306 ✅
@SergioFLS ✅
@abelbriggs1 ✅
@LMFinish ✅
@jimkoutso2008 ✅
@KrisCrossApple

(there's also casriel but the only commit change was a change to add Cinnamon to the README) ✅

`git shortlog` for reference

```
:) mrpowergamerbr@deeparch-whistler:~/Projects/Butterscotch (main)$ git shortlog -sne -- src/
   705  MrPowerGamerBR 
    14  Fancy2209 
     8  Eliandro 
     7  Fancy2209 
     7  Luke Warner
     5  mattx
     4  Krisp
     4  LDA
     3  Classic
     1  Abel Briggs
     1  Kevin Andrade
     1  KrisCrossApple
     1  LDA [48281th GH alt]
     1  Serg One Zero
     1  Serg One Zero
```

If push comes to shove and someone does not agree to the relicense, what will be done is that I will do a reimplementation of their changes while not looking at the original changes, by only looking at the commit title/description to figure out what needs to be reimplemented. ~92% of Butterscotch's commits were made by myself.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,674 +1,373 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
-                            Preamble
-
-  The GNU General Public License is a free, copyleft license for
-software and other kinds of works.
-
-  The licenses for most software and other practical works are designed
-to take away your freedom to share and change the works.  By contrast,
-the GNU General Public License is intended to guarantee your freedom to
-share and change all versions of a program--to make sure it remains free
-software for all its users.  We, the Free Software Foundation, use the
-GNU General Public License for most of our software; it applies also to
-any other work released this way by its authors.  You can apply it to
-your programs, too.
-
-  When we speak of free software, we are referring to freedom, not
-price.  Our General Public Licenses are designed to make sure that you
-have the freedom to distribute copies of free software (and charge for
-them if you wish), that you receive source code or can get it if you
-want it, that you can change the software or use pieces of it in new
-free programs, and that you know you can do these things.
-
-  To protect your rights, we need to prevent others from denying you
-these rights or asking you to surrender the rights.  Therefore, you have
-certain responsibilities if you distribute copies of the software, or if
-you modify it: responsibilities to respect the freedom of others.
-
-  For example, if you distribute copies of such a program, whether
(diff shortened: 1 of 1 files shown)


seija-amanojaku:
I agree

LukeWarnut:
I agree.

jimkoutso2008:
proceed
okay yea jokes aside 
I agree

Eliandro4:
I agree

Classic0306:
I agree even if I don't fully understand it!!!

mattx433:
I agree

Fancy2209:
Proceed.

KSP0O:
**I agree.**

LMFinish:
Go ahead

abelbriggs1:
I agree.

casrielasriel:
"(there's also casriel but the only commit change was a change to add Cinnamon to the README)"

eh well yeah i agree LOL

SergioFLS:
I agree to relicense/dual-license to MPL-2.0 (my changes weren't that huge, anyway)

MrPowerGamerBR:
image

PR #106, opened 2026-05-11
Stub/implement os_is_paused, psn_init_np_libs, and psn_setup_trophies

LukeWarnut:
Addresses multiple missing functions when running the PS Vita version of UNDERTALE with `os_type = psvita`.

`os_is_paused` is used to check if the game has been interrupted by the operating system on various devices. For simplicity, this is always set to return false (unpaused).

`psn_init_np_libs` is stubbed.

`psn_setup_trophies` is used to check if the game has finished setting up Trophies (PlayStation achievements). Previously when running UNDERTALE with `os_type = psvita`, the game would get stuck on a screen waiting for the trophies to get set up. Forcing `psn_setup_trophies` to always return 1 allows us to get past this screen.

### DIFF
--- a/src/vm_builtins.c
+++ b/src/vm_builtins.c
@@ -2599,6 +2599,10 @@ static RValue builtinOsGetRegion(MAYBE_UNUSED VMContext* ctx, MAYBE_UNUSED RValu
     return RValue_makeOwnedString(safeStrdup("US"));
 }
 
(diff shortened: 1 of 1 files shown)


MrPowerGamerBR:
In the future it would be nice to expose the "is paused" as a Runner flag (similar to os type and debug mode)

But for now, this is already good enough :)

PR #113, opened 2026-05-12
SDL 1.2 backend

Un1q32:
SDL 1.2 backend that supports the legacy-gl renderer.  Also has some skeleton support for a software renderer, tho that's not in yet.
Most of the work was done by Fancy.


src/glfw/main.c
```diff
@@ -710,7 +710,6 @@ static void keyCallback(GLFWwindow* window, int key, int scancode, int action, i
     // During playback, suppress real keyboard input (window events like close still work)
     if (InputRecording_isPlaybackActive(globalInputRecording)) return;
     int32_t gmlKey = glfwKeyToGml(key);
-    if (0 > gmlKey) return;
```
Fancy2209, line 713: Why was this removed?

src/sdl/main.c
```diff
@@ -0,0 +1,1219 @@
...
+    }
+
+    fbW = width;
+    fbH = height;
+    nextFb = SDL_CreateRGBSurfaceFrom(
+        framebuffer,
+        width,
+        height,
+        32,
+        width * 4,
```
Fancy2209, line 594: If the buffer has no alpha the BPP can probably be dropped down to 24 and pitch to width * 3

src/glfw/main.c
Un1q32: Theres an equivalent check in RunnerKeyboard_onKey(Up|Down), seemed pointless to check twice.  I can add it back if you want.

src/sdl/main.c
Un1q32: image

Fancy2209: Weird but oh well
Keep as is then

PR #114, opened 2026-05-12
ps3: Add support for paletted textures to reduce VRAM usage

MrPowerGamerBR:
It is still VERY hacky, but it does work


vendor/ps3gl/ps3gl.c
```diff
@@ -162,7 +168,7 @@ void glEnable( GLenum cap )
 			_opengl_state.fog_enabled = true;
 			break;
 		case GL_TEXTURE_2D:
-			_opengl_state.texture0_enabled = true;
+			_opengl_state.texture_unit_enabled[_opengl_state.active_texture_unit] = true;
```
Fancy2209, line 171: Could you use activeBoundTex() here and in line 198 and 229?

vendor/ps3gl/ps3gl.h
```diff
@@ -9,6 +9,46 @@ extern "C"
...
 void ps3glSwapBuffers(void);
 
+// Multi-texture-unit support. Pass GL_TEXTURE0..GL_TEXTURE3.
+void glActiveTexture(GLenum texture);
+
+#define GL_VERTEX_SHADER                0x8B31
+#define GL_FRAGMENT_SHADER              0x8B30
+#define GL_COMPILE_STATUS               0x8B81
+#define GL_LINK_STATUS                  0x8B82
+#define GL_INFO_LOG_LENGTH              0x8B84
```
Fancy2209, line 19: I'd prefer these be appended to gl.h or glext.h, though some of these should already be on those.

Fancy2209: Looking better, these are in those headers already, just need the GL_GLEXT_PROTOTYPES define

src/ps3/main.c
```diff
@@ -25,6 +26,12 @@
 #include "utils.h"
 #include "profiler.h"
 
+// Paletted fragment shader.
+extern unsigned char paletted_fpo[];
```
Fancy2209, line 30: These variable names are super inconsistent?

src/ps3/ps3_textures.c
```diff
@@ -0,0 +1,184 @@
...
+#define PAGE_HEADER_SIZE 12  // u16 w, u16 h, u32 pixelOffset, u32 pixelDataSize
+
+typedef struct {
+    uint16_t width;
+    uint16_t height;
+    uint32_t pixelOffset;
+    uint32_t pixelDataSize;
+} PageInfo;
+
+static FILE*       gFp;
```
Fancy2209, line 18: I could be wrong but do we use gVarName for naming globals anywhere else in butterscotch?

Fancy2209 (approved):
Just minor nitpicks you're free to ignore if you want.

src/ps3/ps3_textures.c
MrPowerGamerBR: It is used for the WASM runner, the global runner variable is `gRunner`

Fancy2209: Okay so we are consistently inconsistent

PR #120, opened 2026-05-14
Re-implement spriteless background layers + partial alpha support

LukeWarnut:
Re-implements the changes in #111 with additional partial support for background colors with alpha levels != 1.

This implementation is an improvement from before, but is not 100% accurate. When using a transparent background, the GMS 2022.9 runner will gradually increase the alpha of the background until it is opaque. Frankly, I'm unsure why the official runner would do this, but this behavior has not been replicated in this PR.

Test program: [game.ios.zip](https://github.com/user-attachments/files/27770704/game.ios.zip)
This program uses one semi-transparent red background and one semi-transparent blue background.

| GMS 2022.9 |  Butterscotch (Before) | Butterscotch (After) |
| -------- | ------- | ------- |
|  | Screenshot 2026-05-14 at 1 56 22 PM |  |

### DIFF
--- a/src/glfw/main.c
+++ b/src/glfw/main.c
@@ -1358,7 +1358,7 @@ int main(int argc, char* argv[]) {
             int gInt = BGR_G(runner->backgroundColor);
             int bInt = BGR_B(runner->backgroundColor);
             int aInt = BGR_A(runner->backgroundColor);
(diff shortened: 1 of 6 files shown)


LukeWarnut:
For reference:
Footage from the official runner using the same background with moving sprites. Notice the trail.  
  
https://github.com/user-attachments/assets/babe6e4b-18fa-489d-bb02-4f07b67fc634

PR #124, opened 2026-05-16
Fix makefile

Un1q32:
I may be stupid, but the guy who designed this language is stupider.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,12 +38,18 @@ ifndef DISABLE_BC17
 DEFINES += -DENABLE_BC17
 endif
 
+# GNU make doesn't have a way to do OR in conditionals, stupid language
 ifndef DISABLE_LEGACY_GL
+ENABLE_GL := 1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -38,13 +38,19 @@ ifndef DISABLE_BC17
 DEFINES += -DENABLE_BC17
 endif
 
+# GNU make doesn't have a way to do OR in conditionals, stupid language for clowns
 ifndef DISABLE_LEGACY_GL
+ENABLE_GL := 1
```
Fancy2209, line 43: Shouldn't this enable Legacy GL and not Modern GL?

Un1q32: ENABLE_GL is just a temp variable im using to avoid having to do an OR (which isn't possible in GNU make), it isnt used anywhere except right below where it's checked to enable compiling gl_common stuff.  I can rename it to something else if you want.

PR #132, opened 2026-05-17
fix stupid typo also move macro to other header

Un1q32:
`__BYTE__ORDER__` is only supposed to have one underscore in the middle.
also moved the macro to another header so it's easy to use in other places (like the software renderer pixel conversion code)


src/endianness.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#if (defined(__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__)) || defined(__BIG_ENDIAN__)
```
Fancy2209: we have common.h for this stuff

Fancy2209 (changes requested):

PR #143, opened 2026-05-19
Fix a couple of egregious warnings

iProgramMC:
This is what's most important.  There are a couple more warnings, such as ones regarding stb_ds usage, and unused parameters and static functions (that actually are something to worry about).

### DIFF
--- a/src/matrix_math.h
+++ b/src/matrix_math.h
@@ -19,7 +19,7 @@ typedef struct {
 
 // ===[ Identity / Copy ]===
 
-static Matrix4f* Matrix4f_identity(Matrix4f* dest) {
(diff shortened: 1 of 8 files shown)


src/matrix_math.h
```diff
@@ -188,9 +188,8 @@ static float Matrix4f_determinant(const Matrix4f *mat) {
 }
 
 // Computes a matrix's inverse and returns true/false if it even exists
-static bool Matrix4f_inverse(Matrix4f *inv, const Matrix4f *mat) {
+static inline inline bool Matrix4f_inverse(Matrix4f *inv, const Matrix4f *mat) {
```
Un1q32: 2 inlines?

iProgramMC: oops

PR #150, opened 2026-05-21
Fix stderr being buffered on windows

Un1q32:
On Windows, SDL 1.2 may call freopen() on stdout and stderr to make them output to files instead of a terminal.  In these cases, stderr might be buffered instead of unbuffered like it should, which means output might be lost if abort() is called.  This setbuf call ensures the stream is always unbuffered, and is effectively a nop if stderr is already unbuffered.

### DIFF
--- a/src/sdl/main.c
+++ b/src/sdl/main.c
@@ -645,6 +645,8 @@ static void onCrashSignal(int sig) {
 // ===[ MAIN ]===
 static bool shouldExit = false;
 int main(int argc, char* argv[]) {
+    setbuf(stderr, NULL);
+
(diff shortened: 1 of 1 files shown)


iProgramMC:
can't we, instead, prevent SDL 1.2 from taking over stderr/stdout somehow?

Un1q32:
> can't we, instead, prevent SDL 1.2 from taking over stderr/stdout somehow?

Doing this requires passing `--disable-stdio-redirect` at configure time for SDL, which might not be possible if someone is using some vendored library.  Also, someone might actually want to redirect stdio for whatever reason.

PR #152, opened 2026-05-22
static-link bzip2

cobaltgit:


cobaltgit:
I'll fix that web build tonight

cobaltgit:
Yay the web build works

Makefile
```diff
@@ -4,18 +4,18 @@
 CC := cc
 
 CFLAGS := -O2 -DNDEBUG
-LIBS := -lbz2
+LIBS :=
```
Un1q32: just remove the whole line

Un1q32:
~~Actually, let me do the Makefile change.  I can do it better.~~

Un1q32:
Actually nevermind nevermind, the idea I had wouldn't work, just keep as is and remove that one line.

PR #160, opened 2026-05-24
Fix building under MSVC

SergioFLS:
Tested on VS2026 Community (latest as of writing this), not sure what happens when using earlier versions

Also ASAN for some reason seems to detect a null pointer reference in dsound.dll, making it crash when compiled with ASAN enabled

This also adds a new dependency exclusive to Windows: [getopt-win32](https://vcpkg.io/en/package/getopt-win32), which has a LGPLv3-only license. Probably not problematic for proprietary forks unless linked statically with it.


src/utils.h
```diff
@@ -20,6 +22,17 @@
 // The "typeof((typeof(n))0" is used to remove the "const" from the typeof
 
 #define repeat(n, it) for (typeof((typeof(n))0) it = 0; it < (n); it++)
+#else
+#define forEach(type, item, array, count) \
```
Un1q32, line 21: for this i think you should have some macro that gets defined to the typeof stuff on gcc and `unsigned long long` on others, so we can duplicate lines less.
Also include `__TINYC__` in the list of compilers compatible with typeof

src/vm.h
```diff
@@ -333,13 +333,17 @@ static inline bool VM_shouldTraceVariable(StringBooleanEntry* traceMap, const ch
...
-    char formatted[strlen(scopeName) + 1 + strlen(varName) + 1];
+    char *formatted = calloc(strlen(scopeName) + 1 + strlen(varName) + 1, 1);
     snprintf(formatted, sizeof(formatted), "%s.%s", scopeName, varName);
-    if (shgeti(traceMap, formatted) != -1) return true;
+    bool match = shgeti(traceMap, formatted) != -1;
+    free(formatted);
+    if (match) return true;
     if (altScopeName != nullptr) {
-        char altFormatted[strlen(altScopeName) + 1 + strlen(varName) + 1];
+        char *altFormatted = calloc(strlen(altScopeName) + 1 + strlen(varName) + 1, 1);
```
Un1q32: i think you should just use malloc() for these, calloc adds overhead by zeroing the memory.  If it was a vla before it was working fine as uninitialized.

SergioFLS: Should be solved in [dce211a](https://github.com/ButterscotchRunner/Butterscotch/pull/160/commits/dce211ad59d36692ab28f1feb467bc088e90bd48)

src/utils.h
SergioFLS: Should be solved in [7b4fb83](https://github.com/ButterscotchRunner/Butterscotch/pull/160/commits/7b4fb83d20bfd4b73282b3bd9a0158339ccde2c8)

src/utils.h
```diff
@@ -10,29 +10,23 @@
...
-
-// The "typeof((typeof(n))0" is used to remove the "const" from the typeof
-
-#define repeat(n, it) for (typeof((typeof(n))0) it = 0; it < (n); it++)
+#if (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 202311L)) \
+    || defined(__GNUC__) || defined(__clang__) || defined(__TINYC__)
+    // The "typeof((typeof(x))0" is used to remove the "const" from the typeof
+    #define TYPEOF(x) typeof((typeof(x))0)
 #else
+    #define TYPEOF(x) long long
```
Un1q32: Do unsigned.

SergioFLS: oops

[3e5bbec](https://github.com/ButterscotchRunner/Butterscotch/pull/160/commits/3e5bbec174519200c33dbb9260fc017356e27e24)

src/utils.h
```diff
@@ -3,23 +3,30 @@
...
+    // The "typeof((typeof(x))0" is used to remove the "const" from the typeof
+    #define TYPEOF(x) typeof((typeof(x))0)
+#else
+    #define TYPEOF(x) unsigned long long
+#endif
+
 #define forEach(type, item, array, count) \
-    for (typeof(count) item##_i_ = 0; item##_i_ < (count); item##_i_++) \
-    for (type* item = &(array)[item##_i_]; item; item = NULL)
+    for (TYPEOF(count) item##_i_ = 0; item##_i_ < (long long)(count); item##_i_++) \
```
Un1q32, line 22: why cast count to long long here?

src/vm_builtins.c
```diff
@@ -1423,7 +1528,8 @@ static RValue builtin_string_length(MAYBE_UNUSED VMContext* ctx, RValue* args, i
 
 // https://docs.vultr.com/clang/examples/remove-all-characters-in-a-string-except-alphabets
 void filterAlphabets(char *str) {
-    char result[strlen(str) + 1];
+	size_t resultLen = strlen(str) + 1;
+    char *result = (char*)malloc(resultLen);
```
Un1q32, line 1532: you should actually use safeMalloc for this so it handles failed allocations right.

src/audio/miniaudio/ma_audio_system.c
```diff
@@ -607,8 +610,10 @@ static void maSetChannelCount(MAYBE_UNUSED AudioSystem* audio, MAYBE_UNUSED int3
 static void maGroupLoad(AudioSystem* audio, int32_t groupIndex) {
     if (groupIndex > 0) {
         int sz = snprintf(nullptr, 0, "audiogroup%d.dat", groupIndex);
-        char buf[sz + 1];
-        snprintf(buf, sizeof(buf), "audiogroup%d.dat", groupIndex);
+		size_t bufSize = (size_t)sz + 1;
+        char *buf = malloc(bufSize);
+        snprintf(buf, bufSize, "audiogroup%d.dat", groupIndex);
+		free(buf);
```
Un1q32, line 616: you free buf here right before it gets used the very next line, you need to free it later

src/profiler.c
```diff
@@ -3,7 +3,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
+#ifndef _MSC_VER
 #include <sys/time.h>
```
Un1q32, line 7: move the `sys/time.h` header to right below `time.h` a little further down.

change it from

```c
#else
#include <time.h>
#endif
```
to
```c
#else
#include <time.h>
#ifndef CLOCK_MONOTONIC
#include <sys/time.h>
#endif
#endif
```

MrPowerGamerBR:
Superseded by https://github.com/ButterscotchRunner/Butterscotch/pull/160

Un1q32:
I have something better for getopt_long, but it's in a branch that requires https://github.com/ButterscotchRunner/Butterscotch/pull/210 to get merged first.

Un1q32:
Additionally, I have a better solution to the alignment thing in that PR.

Un1q32:
This actually seems pretty stale, I might finish this work myself after after the GCC 2.7 and old libc support PRs are in.

SergioFLS:
Because of how stale this PR is, I'm closing it, but at least @Un1q32 is referencing this for other PRs (#233 and #245).

glhf 👍

PR #169, opened 2026-05-27
vm: use switch-case in inline OP_CONV paths

cobaltgit:
Squeezes more juice out of the conv fastpaths


cobaltgit:
Could go even further by refactoring the VM to use computed gotos instead of switch-case

PR #177, opened 2026-05-30
GCC 2.95 support

Un1q32:
This is ready but depends on #176 so its marked as draft until that gets merged.

Adds support for building with GCC 2.95.

GCC 2.95's C mode, even with `-std=gnu9x`, doesn't support mixed declarations and code, which makes it impossible to adapt butterscotch to support it without *major* effort.  It would basically be a full C89 conversion, and would be a monumental diff.  This is unacceptable.
The solution was to target GCC 2.95's C++ mode.  C++ has supported mixed declarations and code for much longer, so I just made the changes required to build with `g++ -fpermissive`.  You must use `-fpermissive` because otherwise GCC will error whenever you implicitly cast a `void *` to another pointer type.

### DIFF
--- a/src/audio/miniaudio/ma_audio_system.c
+++ b/src/audio/miniaudio/ma_audio_system.c
@@ -618,10 +618,9 @@ static void maGroupLoad(AudioSystem* audio, int32_t groupIndex) {
             return;
         }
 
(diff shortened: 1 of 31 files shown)


Un1q32:
This has a fuckton of merge conflicts now but I don't wanna fix it just for it to break again in a couple days.  I'll wait for #176 and then maybe try to do gcc 2.96RH so the diff can be smaller.

Un1q32:
Rebased and ready as of now.

PR #185, opened 2026-06-04
Old GCC workflows + more GCC 2.95 fixes + Makefile improvements

Un1q32:
I didn't catch the bool thing before because I just patched my GCC 2.95's stdbool.h to work in C++ mode and forgot to revert it.


Un1q32:
CI seems to be breaking because it's running the workflow on the upstream branch instead of the PR branch for some reason.  It passes in my repo.

Un1q32:
I just thought of a slightly better way to do this one sec

Un1q32:
Ok now its ready

Un1q32:
bump

Un1q32:
This has been totally redone now.  Also CI is fixed

PR #186, opened 2026-06-04
Add the needed functions for UNDERTALE Yellow

casrielasriel:
Adds the needed functions for Undertale Yellow.
There is a physics system but it is not finished as a few things must be tweaked.


casrielasriel:
i forgot something with box2d, will fix

PR #192, opened 2026-06-05
UNDERTALE Yellow changes

casrielasriel:
Adds the majority of the needed functions for UNDERTALE Yellow.
There is a physics system that needs more configuring to work.


src/desktop/backends/glfw3.c
```diff
@@ -154,16 +154,26 @@ static int32_t glfwMouseButtonToGml(int glfwButton) {
...
+
+    if (winWidth <= 0 || winHeight <= 0) return;
+
+    g_runner->mouse->normalizedX = (xpos - g_runner->viewportX) / g_runner->viewportW;
+    g_runner->mouse->normalizedY = (ypos - g_runner->viewportY) / g_runner->viewportH;
+}
+
 static void mouseButtonCallback(GLFWwindow* window, int button, int action, int mods) {
-    (void)mods; (void)window;
+    (void)mods;
```
Un1q32, line 168: whats the point of this change, the `(void)window` is intended to silence a warning because the window variable doesn't get used in this function.

Un1q32: Oh wait does it cause issues for the global?  In that case rename the function parameter to `_window` and add `(void)_window`

casrielasriel:
It does work and everything is good but it needs to be cleaned up a LOT

PR #193, opened 2026-06-05
SDL3 desktop frontend

cobaltgit:


cobaltgit:
Now with gamepad support (thanks @Eliandro4!)

cobaltgit:
SDL3 now has the most efficient framelimiter according to perf reports..

src/desktop/backends/sdl3.c
```diff
@@ -0,0 +1,454 @@
...
+        SDL_Delay((Uint32)((remaining - 1000000) / 1000000));
+    }
+
+    remaining = time - nowNanos();
+    if (remaining > 0) {
+        SDL_DelayPrecise((Uint64)remaining);
+    }
+}
+
+void platformGamepad_poll(RunnerGamepadState* gp) {
```
Un1q32: Remove this function, it is useless

src/desktop/backends/sdl3.c
```diff
@@ -0,0 +1,454 @@
...
+
+#if defined(ENABLE_MODERN_GL) || defined(ENABLE_LEGACY_GL)
+
+void *platformGetProcAddress(const char *name) {
+    return SDL_GL_GetProcAddress(name);
+}
+
+#endif
+
+double platformGetTime(void) {
```
Un1q32: Remove this function, it is useless after the time refactor.

cobaltgit:
(minor spelling mistake)

Un1q32:
> SDL3 now has the most efficient framelimiter according to perf reports..

If you just use SDL_DelayPrecice for the whole thing does this change? I would imagine DelayPrecice might have logic similar to what we do with sleeping all but the last bit for efficiency.  Making it all one function call would be nice to simplify the code.

Makefile
```diff
@@ -77,6 +77,12 @@ SDL2_LIBS += $(shell pkg-config $(PKG_CONFIG_FLAGS) --libs sdl2)
 LIBS += $(SDL2_LIBS)
 DEFINES += -DUSE_SDL2
 endif
+ifeq ($(DESKTOP_BACKEND),sdl3)
+SDL3_LIBS += $(shell pkg-config --libs sdl3)
```
Un1q32: Add the PKG_CONFIG_FLAGS thing here like the other backends have.

cobaltgit:
> I would imagine DelayPrecice for the whole thing does this change? I would imagine DelayPrecice might have logic similar to what we do with sleeping all but the last bit for efficiency. Making it all one function call would be nice to simplify the code.

Seems like the best call, yeah SDL_DelayPrecise is abit similar to sleeping and then spin-waiting but it's nanosecond-accurate so you don't really need to spin-wait the last <1ms out

PR #197, opened 2026-06-06
implement window_set_cursor and window_get_cursor

Eliandro4:


Un1q32:
Since you're adding the function to the runner struct and not using it outside of that, make the function static and don't add it to the header.  This makes it more clear that it's an optional function and new backends don't need to implement it.

Un1q32:
You should do SDL 3 also since that's in now, it will likely be mostly the same as SDL 2.

Eliandro4:
oh yea, gotta go back to work

Eliandro4:
finished

PR #202, opened 2026-06-07
Make spin-waiting a little more power efficient on desktop

cobaltgit:
Supports x86, ARMv7+, RISC-V GCC/Clang/MSVC, otherwise falls back to a no-op.

Uses the pause/yield instruction to stop busy-looping from using CPU while waiting for the next frame to begin

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -35,3 +35,15 @@
         #define MAYBE_UNUSED
     #endif
 #endif
+
+#if defined(__x86_64__) || defined(__i386__) || defined(_M_X64) || defined(_M_IX86)
+    #include <immintrin.h>
+    #define YIELD() _mm_pause()
+#elif defined(_M_ARM64) || defined(_M_ARM)
+    #include <intrin.h>
+    #define YIELD() __yield()
+#elif defined(__aarch64__) || defined(__arm__) && (__ARM_ARCH >= 7)
+    #define YIELD() __asm__ volatile("yield" ::: "memory")
+#else
+    #define YIELD() ((void)0)
+#endif
--- a/src/common.h
+++ b/src/common.h
@@ -37,13 +37,25 @@
 #endif
 
 #if defined(__x86_64__) || defined(__i386__) || defined(_M_X64) || defined(_M_IX86)
-    #include <immintrin.h>
-    #define YIELD() _mm_pause()
+    #if defined(__GNUC__) && ((__GNUC__ < 4) || (__GNUC__ == 4 && __GNUC_MINOR__ < 4))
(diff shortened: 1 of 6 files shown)


src/common.h
```diff
@@ -35,3 +35,27 @@
...
+        #define YIELD() __asm__ volatile("rep; nop" ::: "memory")
+    #else
+        #include <immintrin.h>
+        #define YIELD() _mm_pause()
+    #endif
+#elif defined(_M_ARM64) || defined(_M_ARM)
+    #include <intrin.h>
+    #define YIELD() __yield()
+#elif defined(__aarch64__) || (defined(__arm__) && defined(__ARM_ARCH) && (__ARM_ARCH >= 7))
+    #if defined(__GNUC__) && (__GNUC__ < 5)
```
Un1q32: clang appears to have this too, clang does define `__GNUC__` but it pretends to be GCC 4.2.1

Un1q32:
Intrinsics are not standard C, so you should put the whole YIELD block behind an `#ifdef __GNUC__`, with a fallback to make it a no-op.

src/common.h
```diff
@@ -35,3 +35,30 @@
         #define MAYBE_UNUSED
     #endif
 #endif
+
+#ifdef __GNUC__
+    #if defined(__x86_64__) || defined(__i386__) || defined(_M_X64) || defined(_M_IX86)
+        #if defined(__GNUC__) && ((__GNUC__ < 4) || (__GNUC__ == 4 && __GNUC_MINOR__ < 4))
```
Un1q32: You can remove the `defined(__GNUC__)` from here and below now since it's guaranteed.

src/common.h
```diff
@@ -35,3 +35,30 @@
...
+
+#ifdef __GNUC__
+    #if defined(__x86_64__) || defined(__i386__) || defined(_M_X64) || defined(_M_IX86)
+        #if ((__GNUC__ < 4) || (__GNUC__ == 4 && __GNUC_MINOR__ < 4))
+            #define YIELD() __asm__ volatile("rep; nop" ::: "memory")
+        #else
+            #include <immintrin.h>
+            #define YIELD() _mm_pause()
+        #endif
+    #elif defined(_M_ARM64) || defined(_M_ARM)
```
Un1q32, line 51: Is this MSVC?  If so make the `#ifdef __GNUC__` into `#if defined(__GNUC__)` and add an `#elif defined(_MSC_VER)` that this can go into.

Un1q32: Also move out the x86 MSVC path.

PR #203, opened 2026-06-08
fix the opengl api error spam

casrielasriel:
top 10 easiest fixes lmao


Un1q32:
Logs are now disabled when NDEBUG is defined, and otherwise enabled.  We don't want to disable logs unconditionally.

PR #204, opened 2026-06-08
HiDPI aware SDL 2

Un1q32:
Makes SDL 2 scale things correctly on HiDPI displays.


Un1q32:
To be clear, it worked before, but on hidpi it would be blurry and suck.

PR #209, opened 2026-06-09
attempt to implement draw_enable_drawevent

cobaltgit:
Is this correct? Hyper princess pitch uses it when transitioning between rooms, screen goes to black for a split second in the GMS port but not in the GM7 original.


Un1q32:
whats up with this

PR #210, opened 2026-06-10
GCC 2.7 support

Un1q32:
I'm still pretty sure we don't need that alignment, but it's here until someone tests the performance impact on a real PS2.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,28 +185,21 @@ all: build/butterscotch
 
 -include $(OBJS:.o=.d)
 
-ifeq ($(filter clean distclean,$(MAKECMDGOALS)),)
-
 -include compat/config.mk
 
 ifndef DISABLE_MMD
 DEPFLAGS = -MMD -MP -MF $(@:.o=.d)
 endif
 
-FORCE:
+$(shell \
+	printf '$(CC)' > compat/tmp/cc-new; \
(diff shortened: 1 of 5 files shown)


Un1q32:
I have another PR that is related but that I've kept separate, that is fully ready but depends on this one, so when this is merged please ping me in the discord so I can immediately put up that PR.

Un1q32:
I've got a loose plan for a kind of roadmap for the changes I plan to make in the future, and this is as good a place to put it as any, so here (subject to change, some of this is likely to happen out of order).

1: this PR

2: legacy/incomplete C library support
- mostly adding shims for systems without all the C99 math functions, also adds a getopt_long shim which will help out MSVC, though even without MSVC support it will still be useful for old BSD systems.

3: "standard C"
- Gating or removing all non-standard extension things (like typeof) so that strictly standards conforming compilers will work.  You can also consider this as "MSVC part 1".

4: remove VLAs
- Removing VLAs, aka "MSVC part 2".  VLAs are bad anyway because we can't guarantee it won't overflow the stack. C++ was right not to add them.

5: Try to look into supporting MSVC in the makefile build
- An actual visual studio solution should be a last resort, it would be a huge maintenance burden.

6: Generally improve the makefile
- Add missing backends like PS2 and PS3, maybe make it's output look nicer.

PR #128, opened 2026-05-17
Add software renderer implementation

iProgramMC:
Taken from https://github.com/iProgramMC/Butterscotch/tree/sw-renderer-old.

This is nowhere near complete. The only game this can run right now is UNDERTALE, and I know of a thing in UNDERTALE that doesn't work also.

PR #217, opened 2026-06-11
add libretro platform

azaka:
Allow the project to be built as a libretro core.

Currently video and input with joystick are tested to work.

Uses software renderer from https://github.com/Un1q32/Butterscotch/tree/software


Un1q32:
PRs that require the software renderer should be marked as draft and wait until #128 is merged.

azaka:
Audio added. Miniaudio is used as mixing engine

leap0x7b:
why software renderer? couldn't opengl be used instead?

Eliandro4:
Maybe its cuz the dev wants the core to run on a device without gou

RicardoFromPortugal:
yeah basically software rendering first, other hw accelerated renderers can be added later down the line

cobaltgit:
Yeah I do think the hardware renderers should be integrated into the core aswell, I (as well as the dev) work with handheld gaming devices that don't have 3D GPUs so it makes sense to integrate the software renderer though

Un1q32:
If you added OpenGL support it could be merged sooner then if it starts as software only, since it wouldn't need the sw renderer to be useful.

azaka:
Core is now hardware accelerated through gles3. Building with software rendering is optional.

Un1q32:
Do we really need to vendor all that library stuff.  Looks *very* big.  Couldn't we just have a workflow that fetches it from wherever you got it from on build time?

azaka:
I have trim down the lib to only include what's needed

cobaltgit:
The desktop backend now detects the opengl version at runtime, reckon libretro should too?

azaka:
The core can request any hardware context that is supported by the frontend and I have tested gles3 so far

cobaltgit:
Oh goodie then when merged with upstream it should work on the miyoo a30 (OpenGL ES 2.0)

src/libretro/libretro.c
```diff
@@ -0,0 +1,733 @@
...
+  glsm_ctl(GLSM_CTL_STATE_CONTEXT_RESET, NULL);
+
+  if (first_init)
+  {
+    glsm_ctl(GLSM_CTL_STATE_SETUP, NULL);
+    first_init = false;
+  }
+#endif
+
+#ifdef ENABLE_GLES
```
Un1q32: We don't use the ENABLE_GLES macro anymore, use ENABLE_MODERN_GL

src/libretro/libretro.c
```diff
@@ -0,0 +1,733 @@
...
+  g_renderer = GLRenderer_create();
+  GLRenderer* gl = (GLRenderer*)g_renderer;
```
Un1q32, line 148: You need to initialize isGLES (a bool in the GLRenderer struct) to something.  Since presumably you're targetting GLES it should be true, but when you want desktop GL instead (which we should support since retroarch has desktop support) it should be false.

Un1q32: In the desktop backend we basically just check glGetString(GL_VERSION) and if it contains "OpenGL ES" then we say we're on GLES.  You should probably do something like that.

cobaltgit:
Oh and it's also hardcoded to link to libGLESv2 etc

cobaltgit:
```
/home/cobalt/Documents/Repos/Butterscotch/src/gl/gl_renderer.c:2355:49: error: implicit declaration of function ‘glUniformMatrix3fv’; did you mean ‘glUniformMatrix4fv’? [-Wimplicit-function-declaration]
 2355 |         else if (uniform.type == GL_FLOAT_MAT3) glUniformMatrix3fv(uniform.location, count / 9, GL_FALSE, values);
      |                                                 ^~~~~~~~~~~~~~~~~~
      |                                                 glUniformMatrix4fv
/home/cobalt/Documents/Repos/Butterscotch/src/gl/gl_renderer.c:2356:49: error: implicit declaration of function ‘glUniformMatrix2fv’; did you mean ‘glUniformMatrix4fv’? [-Wimplicit-function-declaration]
 2356 |         else if (uniform.type == GL_FLOAT_MAT2) glUniformMatrix2fv(uniform.location, count / 4, GL_FALSE, values);
      |                                                 ^~~~~~~~~~~~~~~~~~
      |                                                 glUniformMatrix4fv
/home/cobalt/Documents/Repos/Butterscotch/src/gl/gl_renderer.c: In function ‘glShaderSetUniformI’:
/home/cobalt/Documents/Repos/Butterscotch/src/gl/gl_renderer.c:2370:47: error: implicit declaration of function ‘glUniform3i’; did you mean ‘glUniform4i’? [-Wimplicit-function-declaration]
 2370 |         else if (uniform.type == GL_INT_VEC3) glUniform3i(uniform.location, value1, value2, value3);
      |                                               ^~~~~~~~~~~
      |                                               glUniform4i
```
Having a little trouble compiling a build

azaka:
I probably need to rebase

cobaltgit:
Yeah as well as there being merge conflicts..
(Deltarune on a snes classic would go crazy)

cobaltgit:
Black screen after selecting a chapter in the full version of Deltarune, seems like `game_change` isn't wired up correctly.

cobaltgit:
Rendering seems to be very broken in Deltarune chapter 3 (save select screen is barely recognisable, black screen when creating a file and starting the chapter), wonder why that is

image

Which is odd cause it's using the same modern-gl renderer that our desktop builds are using

cobaltgit:
Also this is based on an older version of the miniaudio backend (tracks don't pitch shift correctly, such as the save select screen music which is meant to be pitched down) etc, as it is now anyway

azaka:
That gfx glitch is probably caused by glsm failing to restore the gl state that butterscotch expects. Or something else that I'm missing. I'll have to get my hands on a copy of chapter 3 to test though

cobaltgit:
Could be, wondering why performance seems worse than the standalone build at least initially.

azaka:
I do notice that issue with recent builds of retroarch. On older versions like 1.7.3 there are no initial slowdowns

cobaltgit:
That's odd, could it be our video settings? What device are you testing RetroArch 1.7.3 on?

azaka:
I've tested RetroArch 1.7.3 on pc.

As for the gfx glitch, renderdoc would be able to pin point the exact cause but I've no luck getting any trace

Un1q32:
This should be rebased on the new backend system, look at the vita and embedded ports for a reference.

Un1q32:
In addition to neeeding a refactor this should prolly not bundle a huge dependency in the way it does.  If it absolutely MUST be included in the source tree then it should live in the vendor folder.

PR #232, opened 2026-06-16
support less complete C libraries

Un1q32:


Un1q32:
This is already complete, I had this ready a while ago and was just waiting on the GCC 2.7 PR to get merged.

PR #234, opened 2026-06-16
Window adjustments and fixes for all backends

AloXado320:
# Main changes for all backends
- Adjust window position and size if a game requests a larger resolution than the current resolution.
- When window size it's changed, it now gets centered if is not maximized and caches the size so it doesn't prevent resize.
- WIN32 time commands are now set in GLFW backends, no need to set on SDL backends since they already do internally.

# Specific backend changes
## GLFW2
- Fixed cursor being locked on Windows when it's invisible.
- Fixed sleep to prevent negative values (Viral Nightmare was affected by it, not other games for some reason).

## SDL1
- Added alternative way to get max resolution on very old versions (GCC 2.95/3.0 requires it).
- Made it so Alt+F4 actually closes the window.

## SDL3
- Ported over HIDPI changes from SDL2.
- Adjusted window Y position so on larger resolutions, the title bar is still visible.

# Description

- Fixes MMX Viral Nightmare, Shotgun Witch and probably other games from setting a larger resolution.
- Fixes Orbinaut Framework games (eg:  Sonic Revisited) setting a large then a small resolution, hiding the title bar.
- Fixes CD-I Remasters from setting the windows size every frame, making them unresizable.

### DIFF
--- a/src/desktop/backends/glfw2.c
+++ b/src/desktop/backends/glfw2.c
@@ -3,6 +3,7 @@
 
 #ifdef _WIN32
 #include <windows.h>
+#include <mmsystem.h>
 #endif
 
 #ifdef ENABLE_SW_RENDERER
@@ -39,9 +40,29 @@ bool platformGetScaledWindowSize(int32_t* outW, int32_t* outH) {
     return platformGetWindowSize(outW, outH);
 }
 
+static void platformGetTrueDesktopSize(int* outW, int* outH) {
+    GLFWvidmode list[256];
+    int numModes = glfwGetVideoModes(list, 256);
+
+    if (numModes > 0) {
+        *outW = list[numModes - 1].Width;
+        *outH = list[numModes - 1].Height;
+    } else {
+        GLFWvidmode mode;
+        glfwGetDesktopMode(&mode);
+        *outW = mode.Width;
+        *outH = mode.Height;
+    }
+}
+
 void platformSetWindowSize(int32_t width, int32_t height) {
     if (width <= 0 || height <= 0) return;
+    PLATFORM_CACHE_WINDOW_SIZE(width, height);
     glfwSetWindowSize(width, height);
+
+    int desktopW, desktopH;
+    platformGetTrueDesktopSize(&desktopW, &desktopH);
+    glfwSetWindowPos(((desktopW - width) / 2), ((desktopH - height) / 2));
 }
 
 void platformGetMousePos(double *xPos, double *yPos) {
@@ -139,8 +160,8 @@ static void GLFWCALL mouseButtonCallback(int button, int action) {
     else if (action == GLFW_RELEASE) RunnerMouse_onButtonUp(g_runner->mouse, gmlButton);
 }
 
-static int g_last_wheel_pos = 0;
 static void GLFWCALL scrollCallback(int pos) {
+    static int g_last_wheel_pos = 0;
     double yoffset = (double)(pos - g_last_wheel_pos);
     g_last_wheel_pos = pos;
     if (g_runner) RunnerMouse_onWheel(g_runner->mouse, yoffset);
@@ -151,7 +172,9 @@ bool platformInit(int32_t reqW, int32_t reqH, const char *title, bool headless) 
         fprintf(stderr, "Headless mode is not supported with GLFW 2\n");
         return false;
     }
-
+#ifdef _WIN32
+    timeBeginPeriod(1);
+#endif
 #ifdef GLFW_OPENGL_VERSION_MAJOR
     if (gfx == SOFTWARE) {
         glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, 1);
@@ -177,6 +200,9 @@ bool platformInit(int32_t reqW, int32_t reqH, const char *title, bool headless) 
     // Init GLFW
     if (!glfwInit()) {
         fprintf(stderr, "Failed to initialize GLFW\n");
+#ifdef _WIN32
+        timeEndPeriod(1);
+#endif
         return false;
     }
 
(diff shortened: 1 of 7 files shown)


src/desktop/backends/sdl1.c
```diff
@@ -59,30 +60,71 @@ static bool platformGetWindowFocus(void) {
     return SDL_GetAppState() & SDL_APPINPUTFOCUS;
 }
 
+// SDL1.2 redirects stdio messages to stdout.txt and stderr.txt on Windows in SDL_main
```
Un1q32: Remove this code entirely, I sometimes want the redirect.  People who don't should build SDL with --disable-stdio-redirect or set the environment variable themselves.

Un1q32: Additionally, `__attribute__((contructor))` requires support from the C library that we can't guarantee, though I think in that case it would just not run and be fine, so it's probably not a big deal.

src/desktop/backends/sdl1.c
```diff
@@ -59,30 +60,71 @@ static bool platformGetWindowFocus(void) {
...
+#if SDL_VERSION_ATLEAST(1, 2, 10)
+    const SDL_VideoInfo* info = SDL_GetVideoInfo();
+    if (info && (reqW >= info->current_w || reqH >= info->current_h)) {
+        PLATFORM_GET_BEST_FIT_RES(reqW, reqH, info->current_w, info->current_h, finalW, finalH);
+        fprintf(stderr, "Warning: Requested resolution %dx%d is bigger than the screen, adjusting to %dx%d\n",
+                reqW, reqH, finalW, finalH);
+    }
+#else
+    // Old SDL1.2: Set a period appropriate resolution as fallback
+    int oldW = 1024;
```
Un1q32: If we ***have*** to have a lowest common denominator resolution to guess it should be 640x480.  I would prefer it if you could find another way to detect the resolution, or just don't run this best fit code on systems with older sdl versions.

src/desktop/platformdefs.h
```diff
@@ -19,6 +19,31 @@ void platformSetWindowSize(int32_t width, int32_t height);
 void platformSetWindowTitle(const char* title);
 void platformSleepUntil(uint64_t time);
 
+// Some games call platformSetWindowSize every frame so this avoids redundant resizing
+#define PLATFORM_CACHE_WINDOW_SIZE(width, height) \
```
Un1q32: Can you just make these `static inline` functions, more readable.

src/desktop/backends/sdl1.c
AloXado320: It's a Windows only thing, though I guess enforcing it it's a bit too much without debugging, alright I'll remove it.

src/desktop/backends/glfw2.c
```diff
@@ -292,7 +346,10 @@ bool platformHandleEvents(void) {
 }
 
 void platformSleepUntil(uint64_t time) {
-    double remaining = ((int64_t)time - nowNanos()) / 1000000000.0;
+    int64_t delta = (int64_t)time - nowNanos();
+    if (delta <= 0) return; // If the target timestamp is already in the past, skip sleeping
```
Un1q32: This seems redundant, what does it fix?

src/desktop/backends/sdl1.c
AloXado320: I choose 1024x768 since that's the common "highest" resolution back then, and it's only applied on first run, though 800x600 would be a good middleground value.

There isn't another way, at least not easily, just look how many backends SDL1 had to change to add the function here libsdl-org/SDL-1.2@d532c73.

src/desktop/platformdefs.h
AloXado320: Sure, should I keep those function in the header or in a new file?

src/desktop/backends/glfw2.c
AloXado320: It fixes MMX Viral Nightmare in GLFW2, without it the game doesn't work and the program hangs.

Un1q32: If you remove it and debug what do you see?  Is `remaining` a super high value?

src/desktop/backends/sdl1.c
Un1q32: 800x600 is fine I guess, people with smaller screens can just resize after the fact.

src/desktop/backends/glfw2.c
AloXado320: Yes, I though it wouldn't happen due to casting but I guess wrapping has it's side effects.

src/desktop/backends/sdl1.c
```diff
@@ -217,21 +264,30 @@ static int32_t SDLMouseButtonToGml(int sdlButton) {
 bool platformHandleEvents(void) {
     SDL_Event e;
     while (SDL_PollEvent(&e)) {
+        switch (e.type) {
+            default:
+                if (InputRecording_isPlaybackActive(globalInputRecording)) continue;
+                break;
```
Un1q32, line 270: remove this break; it does nothing

Un1q32:
This is a very big diff, and it has parts that I know aren't ready.  Could you make a 2nd PR with just the following bits that are pretty simple and look ready

SDL1 ALT+F4 fix
SOME of the SDL3 HIDPI changes, have the new platformGet(Scaled)WindowSize, and remove fbWidth/fbHeight and all that, but don't create a HIDPI window
time(Begin/End)Period changes
GLFW2 platformSleepUntil cast fix thing
All the other small stuff, like tweaks to comments and layout changes that make the code nicer with no functional difference.

AloXado320:
There were window related fixes as the title said, though it makes sense to add some of them separately, the small fixes would be a separate PR, though for the window things it would make more sense to actually implement the window VM builtins (fullscreen, window pos, center, etc).

Would keep it as a draft and take it's changes separately so it can be properly reviewed, the idea would be like this for each PR:

- Backend fixes only (glfw2 sleep/mouse, glfw time functions, sdl1 alt+f4)
- SDL3 HIDPI
- Window VM implementation for each backend

The large resoultion fix could be in the window vm pr or after it was reviewed or merged (alongside support for old sdl1.2/sdl2.0).

PR #242, opened 2026-06-18
Changed 2.2+ to 2.0+

SplaatSites:
Changed 2.2+ to 2.0+ because of the March 2019 demo being built on GameMaker: Studio 2.0.0.0. https://github.com/ButterscotchRunner/CompatibilityList/issues/126


MrPowerGamerBR:
Earlier GameMaker: Studio 2 versions use WAD Version 16 (see SURVEY_PROGRAM for reference)

PR #245, opened 2026-06-19
Explicit pointer casts

Un1q32:
Adds explicit pointer casts everywhere.
Also adds a C++ workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -390,7 +390,7 @@ jobs:
             flags: '-std=gnu99 -w'
           - name: GCC 2.95
             compiler: g++-2.95
-            flags: '-fpermissive -w'
+            flags: '-w'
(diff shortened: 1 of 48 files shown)


Un1q32:
This is ready once #233 is in.

Un1q32:
My plan for this PR now is to make separate PRs with all the bits that aren't the explicit pointer casts, because the explicit pointer casts are most of the diff.
Explicit pointer casts also have the benefit of not changing codegen so I can just build before and after and if the binary hash doesn't change then it's almost certainly not breaking anything accidentally.

Un1q32:
This is just explicit pointer casts now, the binary hash is identical before and after so codegen definitely isn't changed.
I'll leave this open for a day so if anyone wants to make sure there's nothing wrong with the workflow or anything, and if nobody objects to anything I'll merge.

PR #252, opened 2026-06-22
Make the modern GL renderer compatible with OpenGL 2.0/OpenGL ES 2.0

cobaltgit:
clanker assisted but hacky


cobaltgit:
Ok shaders are broken on gles but work on desktop...send help

cobaltgit:
More specifically shaders don't work on SDL..

cobaltgit:
Shaders now *should* work in gles 2.0

cobaltgit:
human.. i remember you're semicolons

src/desktop/backends/glfw3.c
```diff
@@ -225,8 +225,23 @@ bool platformInit(int32_t reqW, int32_t reqH, const char *title, bool headless)
...
+#else
+        glfwWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_API);
+        glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 2);
+        glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 1);
+#endif
+        window = glfwCreateWindow(reqW, reqH, title, nullptr, nullptr);
+    }
     if (!window) {
-        fprintf(stderr, "Failed to create GLFW window\n");
+        fprintf(stderr, "Failed to create GLFW window )\n");
```
Un1q32: typo?

src/desktop/backends/sdl2.c
```diff
@@ -95,15 +95,13 @@ bool platformInit(int reqW, int reqH, const char *title, bool headless) {
...
-#ifdef SDL_GL_CONTEXT_PROFILE_MASK
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES);
-#endif
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
 #else
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
         SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
-        SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
+        SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_COMPATIBILITY);
```
Un1q32: I think compatibility is the default, you dont need to explicitly set it

cobaltgit:
Ok we need to find a way to pass down the gles runtime check to the platforms

src/desktop/backends/sdl2.c
```diff
@@ -14,6 +14,61 @@ static Runner *g_runner;
 static SDL_Surface* scr;
 static SDL_Window *window;
 
+static SDL_Window *tryOpenWindow(int reqW, int reqH, const char* title, Uint32 flags, SDL_GLContext* out_context) {
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {        
+        SDL_Window *newWindow;
+        int contextFlags = 0;
+
+#if SDL_VERSION_ATLEAST(2, 0, 2)
```
Un1q32: instead of a version check, just always do the old version path, it's equivalent so the version check just adds more lines

src/desktop/backends/sdl2.c
```diff
@@ -14,6 +14,53 @@ static Runner *g_runner;
 static SDL_Surface* scr;
 static SDL_Window *window;
 
+static SDL_Window *tryOpenWindow(int reqW, int reqH, const char* title, Uint32 flags, SDL_GLContext* out_context) {
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {        
+        SDL_Window *newWindow;
+        int contextFlags = SDL_GL_CONTEXT_DEBUG_FLAG;
```
Un1q32: Only set the debug flag if NDEBUG isn't defined, else init this to 0

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,58 @@
 
 static Runner *g_runner;
 
+static bool tryOpenWindow(int reqW, int reqH) {
+    if (gfx == SOFTWARE || gfx == LEGACY_GL) {
+#ifdef GLFW_OPENGL_VERSION_MAJOR
```
Un1q32, line 22: All these macros were added at the same time, wrap the whole function in 
```c
#ifdef GLFW_OPENGL_VERSION_MAJOR
// the whole tryOpenWindow function body
#else
    return glfwOpenWindow(reqW, reqH, 8, 8, 8, 8, 24, 8, GLFW_WINDOW);
#endif
```

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,58 @@
...
+                glfwOpenWindowHint(GLFW_OPENGL_PROFILE, 0);
+            }
+#endif
+
+#ifdef GLFW_OPENGL_DEBUG_CONTEXT
+            glfwOpenWindowHint(GLFW_OPENGL_DEBUG_CONTEXT, GL_TRUE);
+#endif
+        }
+#endif
+        if (glfwOpenWindow(reqW, reqH, 8, 8, 8, 8, 24, 8, GLFW_WINDOW) == GL_TRUE) {
```
Un1q32: remove the `== GL_TRUE` and just check if non-zero.

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,58 @@
...
+            return true;
+        }
+        glfwCloseWindow();
+        fprintf(stderr, "Failed to create OpenGL %s %d.%d context, retrying with next version...\n", GLCommon_versions[i].gles ? "ES" : "Core", GLCommon_versions[i].major, GLCommon_versions[i].minor);
```
Un1q32: Don't log fails like this

src/gl/gl_renderer.c
```diff
@@ -251,7 +293,100 @@ static void glInit(Renderer* renderer, DataWin* dataWin) {
     renderer->dataWin = dataWin;
 
     GMLShader* defaultShader = safeCalloc(1, sizeof(GMLShader));
-    bool success = compileProgram(defaultShader, "default", defaultVertexShaderSource, defaultFragmentShaderSource, 0, nullptr);
+    const char* versionStr = (const char*) glGetString(GL_VERSION);
+    fprintf(stderr, "GL: versionStr=%s\n", versionStr);
```
Un1q32: change this from `GL: versionStr=%s\n` to `OpenGL version: %s\n` to be less ugly

src/gl/gl_renderer.c
```diff
@@ -251,7 +293,100 @@ static void glInit(Renderer* renderer, DataWin* dataWin) {
...
+        }
+    }
+    
+    gl->isGL3 = (major >= 3);
+    
+    const char* extensions = (const char*) glGetString(GL_EXTENSIONS);
+    gl->hasVAO = gl->isGL3 || (extensions != nullptr && (strstr(extensions, "GL_ARB_vertex_array_object") || \
+            strstr(extensions, "GL_OES_vertex_array_object")));
+    
+    bool hasFBO =
```
Un1q32: instead of this just do `bool hasFBO = (glWhateverFBO || glWhateverFBOARB || glWhateverFBOEXT || glWhateverFBOOES)`
glad has done the checks for us, string operations are wasteful and ugly.

cobaltgit: But won't those always return true since they're defined at the top

Un1q32: Oh actually you're right. disregard that suggestion and revert the last commit.

src/gl/gl_renderer.c
```diff
@@ -26,51 +26,86 @@
...
+static void rt_glGenVertexArrays(GLsizei n, GLuint* arrays) {
+    if (glGenVertexArrays) glGenVertexArrays(n, arrays);
+    else if (glGenVertexArraysOES) glGenVertexArraysOES(n, arrays);
+}
+#undef glGenVertexArrays
+#define glGenVertexArrays rt_glGenVertexArrays
+
+static void rt_glDeleteVertexArrays(GLsizei n, const GLuint* arrays) {
+    if (glDeleteVertexArrays) glDeleteVertexArrays(n, arrays);
+    else if (glDeleteVertexArraysOES) glDeleteVertexArraysOES(n, arrays);
```
Un1q32: Aren't you guarding all these functions behind bools anyway?  If so you should replace all these if elses with just plain else.  That way if you forgot to guard one of the calls it will segfault instead of potentially hiding a bug.

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,55 @@
...
+        return glfwOpenWindow(reqW, reqH, 8, 8, 8, 8, 24, 8, GLFW_WINDOW) != 0;
+    }
+
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, GLCommon_versions[i].major);
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, GLCommon_versions[i].minor);
+
+        if (GLCommon_versions[i].gles) {
+#ifdef GLFW_OPENGL_ES
```
Un1q32: remove this ifdef, no longer needed

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,55 @@
...
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, GLCommon_versions[i].major);
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, GLCommon_versions[i].minor);
+
+        if (GLCommon_versions[i].gles) {
+#ifdef GLFW_OPENGL_ES 
+            glfwOpenWindowHint(GLFW_OPENGL_ES, GL_TRUE);
+#endif
+        } else {
+#ifdef GLFW_OPENGL_ES
```
Un1q32: this one too

src/desktop/backends/sdl2.c
```diff
@@ -128,6 +160,20 @@ bool platformInit(int reqW, int reqH, const char *title, bool headless) {
             reqW, reqH,
             flags
     );
+    
+    SDL_GLContext gl_context = NULL;
+    
+    if (window && gfx != SOFTWARE) {
+        gl_context = SDL_GL_CreateContext(window);
```
Un1q32: make this glContext instead of gl_context

src/desktop/backends/sdl3.c
```diff
@@ -37,6 +37,53 @@ static const int SDL_TO_GML_BUTTON[SDL_GAMEPAD_BUTTON_COUNT] = {
...
+
+        newWindow = SDL_CreateWindow(
+            title,
+            reqW,
+            reqH,
+            flags
+        );
+
+        if (newWindow) {
+            *out_context = SDL_GL_CreateContext(newWindow);
```
Un1q32: camelCase variable

src/desktop/backends/glfw2.c
cobaltgit:
```
src/desktop/backends/glfw2.c:35:32: error: 'GLFW_OPENGL_ES' undeclared (first use in this function); did you mean 'GLFW_OPENED'?
   35 |             glfwOpenWindowHint(GLFW_OPENGL_ES, GL_TRUE);
      |                                ^~~~~~~~~~~~~~
      |                                GLFW_OPENED
```
Oh (windows glfw2)

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,51 @@
...
+        return glfwOpenWindow(reqW, reqH, 8, 8, 8, 8, 24, 8, GLFW_WINDOW) != 0;
+    }
+
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, GLCommon_versions[i].major);
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, GLCommon_versions[i].minor);
+
+        if (GLCommon_versions[i].gles) {
+            glfwOpenWindowHint(GLFW_OPENGL_ES, GL_TRUE);
```
Un1q32: shouldn't this be `glfwOpenWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API);`

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,51 @@
...
+        return glfwOpenWindow(reqW, reqH, 8, 8, 8, 8, 24, 8, GLFW_WINDOW) != 0;
+    }
+
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, GLCommon_versions[i].major);
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, GLCommon_versions[i].minor);
+
+        if (GLCommon_versions[i].gles) {
+            glfwOpenWindowHint(GLFW_CLIENT_API, GLFW_OPENGL_ES_API);
```
Un1q32: Ok yea put the #ifdef back.  Also put the GLES entries in the struct in platformdefs.h behind
`#if !defined(USE_GLFW2) || defined(GLFW_OPENGL_ES_API)`

cobaltgit: They're already in platformdefs though right?

Un1q32: ?

Un1q32: Maybe I explained badly, one sec.

src/desktop/platformdefs.h
```diff
@@ -26,5 +26,32 @@ enum GraphicsAPI {
...
+    { 4, 1, false },
+    { 4, 0, false },
+    { 3, 3, false },
+    { 3, 2, false },
+    { 3, 1, false },
+    { 3, 0, false },
+    { 2, 1, false },
+    { 2, 0, false },
+    /* GLES */
+    { 3, 2, true  },
```
Un1q32, line 54: add `#if !defined(USE_GLFW2) || defined(GLFW_OPENGL_ES_API)` right here to guard all the GLES entries.  That way glfw2 versions that don't support GLES won't try to use them.

Un1q32: Oh, actually glfw2 on linux appears to not support GLES, so it's prolly not supported anywhere with glfw2, so just do `#ifndef USE_GLFW2`

cobaltgit: OH sorry i'm being stupid atm

src/desktop/backends/glfw2.c
```diff
@@ -18,6 +18,51 @@
...
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, (gfx == SOFTWARE) ? 0 : 1);
+        return glfwOpenWindow(reqW, reqH, 8, 8, 8, 8, 24, 8, GLFW_WINDOW) != 0;
+    }
+
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MAJOR, GLCommon_versions[i].major);
+        glfwOpenWindowHint(GLFW_OPENGL_VERSION_MINOR, GLCommon_versions[i].minor);
+
+        if (GLCommon_versions[i].gles) {
```
Un1q32: remove this if branch entirely, since glfw2 appears to not support gles at all.

src/desktop/backends/sdl2.c
```diff
@@ -149,14 +195,14 @@ bool platformInit(int reqW, int reqH, const char *title, bool headless) {
         return false;
     }
     if (gfx != SOFTWARE) {
-        if (!SDL_GL_CreateContext(window)) {
+        if (!gl_context) {
```
Un1q32: use glContext here

src/desktop/backends/sdl2.c
```diff
@@ -128,6 +160,20 @@ bool platformInit(int reqW, int reqH, const char *title, bool headless) {
             reqW, reqH,
             flags
     );
+
```
Un1q32, line 181: You should remove the above call to SDL_CreateWindow and only use tryOpenWindow for window creation.
Currently opening a window in macOS gives a GL 2.1 context, when we can do a lot better than that.
Do the same to SDL 3 as well.

cobaltgit:
Gotta fix legacy-gl

src/desktop/backends/sdl2.c
```diff
@@ -149,14 +215,14 @@ bool platformInit(int reqW, int reqH, const char *title, bool headless) {
         return false;
     }
     if (gfx != SOFTWARE) {
-        if (!SDL_GL_CreateContext(window)) {
+        if (!glContext) {
```
Un1q32: Just remove this check, since tryOpenWindow will fail if it can't create a context
Same for SDL 3

src/desktop/backends/sdl2.c
```diff
@@ -14,6 +14,88 @@ static Runner *g_runner;
 static SDL_Surface* scr;
 static SDL_Window *window;
 
+static SDL_Window *tryOpenWindow(int reqW, int reqH, const char* title, Uint32 flags, SDL_GLContext* outContext) {
```
Un1q32: We don't use the SDL_GLContext for anything after creation, so you don't need to return it like this.
Same is true for SDL 3

src/desktop/backends/sdl3.c
```diff
@@ -37,6 +37,82 @@ static const int SDL_TO_GML_BUTTON[SDL_GAMEPAD_BUTTON_COUNT] = {
...
+        if (newWindow) {
+            if (SDL_GL_CreateContext(newWindow)) {
+                return newWindow;
+            }
+            SDL_DestroyWindow(newWindow);
+        }
+        return NULL;
+    }
+    int i;
+    for (i = 0; i < (int)(sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0])); i++) {
```
Un1q32: Declare i inside the loop, and make it size_t, and don't cast the sizeof divide thing.

Un1q32: Also do the same for the other backends.

src/gl/gl_renderer.c
```diff
@@ -251,7 +293,100 @@ static void glInit(Renderer* renderer, DataWin* dataWin) {
...
+            "%s"
+            "varying vec2 vTexCoord;\nvarying vec4 vColor;\n"
+            "#define TEXTURE_2D texture2D\n#define FRAG_COLOR gl_FragColor\n%s", 
+            fragHeader, baseFragmentShader);
+    }
+
+    const char* defaultAttributes[] = { "aPos", "aColor", "aTexCoord" };
+    bool success = compileProgram(defaultShader, "default", vertSrc, fragSrc, 3, defaultAttributes);
     if (!success) {
         fprintf(stderr, "GL: Failed to compile default shaders! Bailing...");
```
Un1q32: This should prolly have a newline at the end btw.

src/gl/gl_renderer.c
```diff
@@ -26,51 +26,102 @@
...
+    "    vec4 c = TEXTURE_2D(uTexture, vTexCoord) * vColor;\n"
+    "    if (uAlphaTestEnabled) {\n"
+    "        if (uAlphaTestRef >= c.a) discard;\n"
+    "    }\n"
     "    c.rgb = mix(c.rgb, uFogColor.rgb, uFogColor.a);\n"
-    "    fragColor = c;\n"
+    "    FRAG_COLOR = c;\n"
     "}\n";
 
+#ifndef GL_FRAMEBUFFER
```
Un1q32: What is this for?  WASM?

src/desktop/backends/sdl2.c
```diff
@@ -13,6 +13,88 @@ static Runner *g_runner;
...
+            if (SDL_GL_CreateContext(newWindow)) {
+                return newWindow;
+            }
+            SDL_DestroyWindow(newWindow);
+        }
+        return NULL;
+    }
+    for (size_t i = 0; i < sizeof(GLCommon_versions)/sizeof(GLCommon_versions[0]); i++) {        
+        SDL_Window *newWindow;
+#ifdef __APPLE__
```
Un1q32: Won't it work without this?

Un1q32 (approved):
LGTM

PR #253, opened 2026-06-23
fix clang warning in stb_vorbis

Un1q32:
taken from https://github.com/nothings/stb/pull/1746
Mostly relevant for the makefile build, which doesn't have an easy way to disable warnings on certain files.
This is also technically a bugfix, since the overflow check was getting optimized away before and now will work, but It was unlikely enough to be an issue that I doubt anyone will notice.


Un1q32:
This is probably fine since it's taken from another PR upstream that used the solution directly suggested by the main project owner.  Said upstream PR has no responses .  I'm sick and tired of this warning and it would be basically impossible to *properly* review this without like, learning the whole ass stb_vorbis codebase, which nobody is gonna do.

PR #254, opened 2026-06-24
Remove bzip2 compression functions

cobaltgit:


CMakeLists.txt
```diff
@@ -150,16 +150,16 @@ target_include_directories(butterscotch PUBLIC vendor/base64)
...
     vendor/bzip2/bzlib.c
-    vendor/bzip2/compress.c
     vendor/bzip2/crctable.c
     vendor/bzip2/decompress.c
     vendor/bzip2/huffman.c
     vendor/bzip2/randtable.c
 )
 target_include_directories(bzip2 PUBLIC vendor/bzip2)
 target_compile_options(bzip2 PRIVATE -w)
+target_compile_definitions(bzip2 PRIVATE BZ_NO_STDIO)
```
Un1q32: instead of doing this, how about just putting `#define BZ_NO_STDIO` at the top of bzlib_private.h

vendor/bzip2/bzlib.c
```diff
@@ -360,127 +305,21 @@ Bool copy_output_until_stop ( EState* s )
...
-         else
-         if (s->strm->avail_in == 0) {
-            break;
-         }
-      }
-
-   }
-
-   return progress_in || progress_out;
+    return False;
```
Un1q32: Instead of stubbing the functions we don't need, remove them entirely and remove their declarations from the headers.  That way if we accidentally try to use a removed function it will error at compile time instead of silently compiling with a bug.

Un1q32:
LGTM

Un1q32 (approved):

PR #255, opened 2026-06-24
Remove all VLAs

Un1q32:
Removes all variable length arrays from the codebase
Predecessor to #245 
We shouldn't have VLAs anyway even if we don't care about C++ compatibility, there's no way to check if an allocation fails.


cobaltgit:
Looks alright to me

PR #257, opened 2026-06-25
All the non-explicit-pointer-cast bits of standard C++

Un1q32:


cobaltgit (approved):
cobalt official stamp of approval

PR #260, opened 2026-06-26
banish pragma once

Un1q32:
The last non-standard C99 thing we require.

Done automatically with a script.


Un1q32:
This was done via script, and I looked at all the files and it seems perfect.  I can't think of a reason this could even theoretically cause any issues.

PR #261, opened 2026-06-26
safe allocation functions

Un1q32:
Old TCC needs this.

### DIFF
--- a/src/utils.h
+++ b/src/utils.h
@@ -12,6 +12,10 @@
 
 #include "real_type.h"
 
+#ifdef PLATFORM_PS2
+#include <malloc.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


Un1q32:
Old TCC really isn't that important, but this will also allow the compiler to make better size/speed tradeoff optimization, since with the macro only version it's really hard for the compiler to know that the same code repeats over and over (although it only reduces binary size by about 16k or so).

Un1q32:
Actually, this also appears to be needed for old MSVC.

Un1q32:
And new MSVC, ok.

cobaltgit (approved):
Looks good to me

PR #264, opened 2026-06-27
Fullscreen

Purplesmaug96:
SDL1 is stubbed
GLFW, SDL2, and SDL3 should now be able to fullscreen
borderless fullscreen builtin functions are registered to normal fullscreen functions


Un1q32:
Rather than making the functions non-static and in platformdefs.h, and putting them in the runner struct in desktop/main.c, make them static, remove from platformdefs.h, and init the runner struct entries in platformInitFunctions.

This makes it clear that these are optional functions and backends don't need to implement them, you can also remove the stub functions from platforms where real fullscreen is impossible like SDL 1.2.

Un1q32:
Also, if it's not too much trouble, could you try and research what the yoyo runner does on platforms where non-fullscreen is impossible like consoles or mobile.

src/vm_builtins.c
```diff
@@ -7596,8 +7596,29 @@ static RValue builtin_joystick_axes(VMContext* ctx, RValue* args, MAYBE_UNUSED i
 }
 
 // Window stubs
-STUB_RETURN_ZERO(window_get_fullscreen)
-STUB_RETURN_UNDEFINED(window_set_fullscreen)
+static RValue builtin_window_get_fullscreen(VMContext* ctx, MAYBE_UNUSED RValue* args, MAYBE_UNUSED int32_t argCount) {
+    Runner* runner = ctx->runner;
+    if (runner != nullptr && runner->getWindowSize != nullptr) {
+        bool fullscreen = false;
+        if (runner->getWindowFullscreen(&fullscreen)) {
```
Un1q32: Why not just make getWindowFullscreen take no arguments and return bool, and use the return value as the result instead of the bool* output thing.  if we're assuming false when failing to get the fullscreen value anyway it would be literally identical.

Purplesmaug96: done

src/desktop/main.c
```diff
@@ -1366,11 +1366,11 @@ int main(int argc, char* argv[]) {
...
         runner->debugMode = args.debug;
         runner->osType = args.osType;
-        runner->setWindowTitle = platformSetWindowTitle;
-        runner->getWindowSize = platformGetWindowSize;
         runner->setWindowSize = platformSetWindowSize;
-		runner->getWindowFullscreen = platformGetWindowFullscreen;
-		runner->setWindowFullscreen = platformSetWindowFullscreen;
+        runner->getWindowSize = platformGetWindowSize;
+        runner->setWindowTitle = platformSetWindowTitle;
+		runner->getWindowFullscreen = nullptr;
```
Un1q32: dont do this, they're already null by default

src/desktop/backends/glfw3.c
```diff
@@ -106,6 +106,14 @@ bool platformGetScaledWindowSize(int32_t* outW, int32_t* outH) {
     return true;
 }
 
+bool platformGetWindowFullscreen(bool* outFullscreen) {
```
Un1q32: You missed this one

src/vm_builtins.c
```diff
@@ -7596,8 +7596,28 @@ static RValue builtin_joystick_axes(VMContext* ctx, RValue* args, MAYBE_UNUSED i
...
-STUB_RETURN_ZERO(window_get_fullscreen)
-STUB_RETURN_UNDEFINED(window_set_fullscreen)
+static RValue builtin_window_get_fullscreen(VMContext* ctx, MAYBE_UNUSED RValue* args, MAYBE_UNUSED int32_t argCount) {
+    Runner* runner = ctx->runner;
+    if (runner != nullptr && runner->getWindowSize != nullptr) {
+        bool fullscreen = runner->getWindowFullscreen();
+        return RValue_makeBool((GMLReal) fullscreen);
+
+    }
+    return RValue_makeReal((GMLReal) false);
```
Un1q32: For this, assume true instead of false.  Most backends that don't define this will be stuff like consoles where fullscreen == true is more accurate.

src/vm_builtins.c
```diff
@@ -7596,8 +7596,28 @@ static RValue builtin_joystick_axes(VMContext* ctx, RValue* args, MAYBE_UNUSED i
 }
 
 // Window stubs
-STUB_RETURN_ZERO(window_get_fullscreen)
-STUB_RETURN_UNDEFINED(window_set_fullscreen)
+static RValue builtin_window_get_fullscreen(VMContext* ctx, MAYBE_UNUSED RValue* args, MAYBE_UNUSED int32_t argCount) {
+    Runner* runner = ctx->runner;
+    if (runner != nullptr && runner->getWindowSize != nullptr) {
+        bool fullscreen = runner->getWindowFullscreen();
+        return RValue_makeBool((GMLReal) fullscreen);
```
Un1q32: just do `return RValue_makeBool(runner->getWindowFullscreen());` on one line

src/vm_builtins.c
Un1q32: oh, also this should be RValue_makeBool instead of RValue_makeReal, and shouldn't cast false to GMLReal.

src/vm_builtins.c
```diff
@@ -7596,8 +7596,28 @@ static RValue builtin_joystick_axes(VMContext* ctx, RValue* args, MAYBE_UNUSED i
...
+    return RValue_makeReal((GMLReal) false);
+}
+
+static RValue builtin_window_set_fullscreen(VMContext* ctx, RValue* args, MAYBE_UNUSED int32_t argCount) {
+    if (argCount > 1) return RValue_makeUndefined();
+
+    Runner* runner = ctx->runner;
+    if (runner != nullptr && runner->setWindowFullscreen != nullptr) {
+		bool fullscreen = RValue_toBool(args[0]);
+		runner->setWindowFullscreen(fullscreen);
```
Un1q32: just do `runner->setWindowFullscreen(RValue_toBool(args[0]));` on one line

src/desktop/main.c
Purplesmaug96: i did that since i thought it'd fix sdl1 from segfaulting when getWindowFullscreen is called (which i then forgot to test) because maybe its filled with garbage memory idk, which wasnt the problem, i forgot to update the nullptr check in builtin_window_get_fullscreen

src/desktop/backends/sdl2.c
```diff
@@ -178,14 +227,15 @@ bool platformInit(int reqW, int reqH, const char *title, bool headless) {
 #if SDL_VERSION_ATLEAST(2, 0, 1)
     flags |= SDL_WINDOW_ALLOW_HIGHDPI;
 #endif
-    
+
+    gFullscreen = false;
```
Un1q32: it will be false by default, but you can be more explicit by initializing the global if you want.

src/vm_builtins.c
```diff
@@ -7596,8 +7596,26 @@ static RValue builtin_joystick_axes(VMContext* ctx, RValue* args, MAYBE_UNUSED i
 }
 
 // Window stubs
-STUB_RETURN_ZERO(window_get_fullscreen)
-STUB_RETURN_UNDEFINED(window_set_fullscreen)
+static RValue builtin_window_get_fullscreen(VMContext* ctx, MAYBE_UNUSED RValue* args, MAYBE_UNUSED int32_t argCount) {
+    Runner* runner = ctx->runner;
+    if (runner != nullptr && runner->getWindowFullscreen != nullptr) {
+        return RValue_makeBool(runner->getWindowFullscreen());
+
```
Un1q32: remove this extra blank line, and add one after the curly brace

Purplesmaug96: this what you mean?

```
static RValue builtin_window_get_fullscreen(VMContext* ctx, MAYBE_UNUSED RValue* args, MAYBE_UNUSED int32_t argCount) {
    Runner* runner = ctx->runner;
    if (runner != nullptr && runner->getWindowFullscreen != nullptr) {
        return RValue_makeBool(runner->getWindowFullscreen());
    }
	
    return RValue_makeBool(true);
}
```

src/desktop/backends/sdl2.c
Purplesmaug96: going to make it false at definition (static bool gFullscreen = false;)

Un1q32: What I mean is, remove this line.

src/runner.h
```diff
@@ -499,6 +499,8 @@ struct Runner {
     void (*setWindowTitle)(const char* title);
     bool (*getWindowSize)(int32_t* outW, int32_t* outH);
     void (*setWindowSize)(int32_t width, int32_t height);
+	bool (*getWindowFullscreen)();
```
Un1q32:
```suggestion
	bool (*getWindowFullscreen)(void);
```
Clang will warn about this with -pedantic in C23 mode.

src/desktop/backends/sdl3.c
```diff
@@ -134,6 +139,11 @@ void platformSetWindowSize(int32_t width, int32_t height) {
         scr = SDL_GetWindowSurface(window);
 }
 
+static void platformSetWindowFullscreen(bool fullscreen) {
+	if (!window) return;
+	gFullscreen = SDL_SetWindowFullscreen(window, fullscreen) ? fullscreen : gFullscreen; // Assign "fullscreen" if success, "gFullscreen" on failure to not modify
```
Un1q32:
```suggestion
    if (SDL_SetWindowFullscreen(window, fullscreen))
	    gFullscreen = fullscreen;
```
Just do this, it's more readable.

src/desktop/backends/glfw3.c
```diff
@@ -116,6 +122,31 @@ void platformSetWindowSize(int32_t width, int32_t height) {
...
+	int32_t h = 0;
+	platformGetWindowSize(&w, &h);
+
+	int refreshRate = glfwGetVideoMode(glfwGetPrimaryMonitor())->refreshRate;
+
+	if (fullscreen) {
+		glfwGetWindowPos(window, &savedWindowX, &savedWindowY);
+		glfwSetWindowMonitor(window, glfwGetPrimaryMonitor(), 0, 0, w, h, refreshRate);
+	}
+	else {
```
Un1q32: move the `else {` up to be on the same line as the `}`

src/desktop/backends/glfw3.c
```diff
@@ -106,6 +106,12 @@ bool platformGetScaledWindowSize(int32_t* outW, int32_t* outH) {
     return true;
 }
 
+static bool platformGetWindowFullscreen() {
+	if (!window) return false;
+
```
Un1q32: remove this blank line

src/runner.h
Purplesmaug96: done

src/desktop/backends/sdl3.c
Purplesmaug96: done

src/desktop/backends/glfw3.c
Purplesmaug96: done

src/desktop/backends/glfw3.c
Purplesmaug96: done

src/desktop/backends/glfw3.c
```diff
@@ -116,6 +121,30 @@ void platformSetWindowSize(int32_t width, int32_t height) {
...
+		glfwSetWindowMonitor(window, glfwGetPrimaryMonitor(), 0, 0, w, h, refreshRate);
+	} else {
+		bool fullscreen = false;
+		platformGetWindowFullscreen(&fullscreen);
+		if (fullscreen) {
+			glfwGetWindowPos(window, &savedWindowX, &savedWindowY);
+		}
+		glfwSetWindowMonitor(window, nullptr, savedWindowX, savedWindowY, w, h, refreshRate);
+	}
+
```
Un1q32: remove this blank line

Purplesmaug96: done

PR #268, opened 2026-06-27
MSVC

Un1q32:
I didn't feel like fucking with cmake so I put msvc build support in the fcking makefile, which is *super* cursed but prolly still makes shit easier than cmake on old systems.


Purplesmaug96:
i did feel like fucking with cmake so i put msvc support in for cmake for my 360 port, feel free to take that
it is largely ai generated though

cobaltgit (approved):

Un1q32:
I'll do cmake later in another PR.

PR #279, opened 2026-06-29
 [BSD, other UNIX-like] Fix audio latency on systems that use OSS

Renkoto:
Tested on FreeBSD 15.1 RELEASE with both Deltarune and Undertale, and with the SDL3 and GLFW 3 when building with the miniaudio backend.

This fix a massive audio latency issue with Butterscotch on *NIX-like systems

I'm unaware if this will cause regressions on ancient FreeBSD versions like 2.2.8 or very old pre-ALSA Linux distros, feel free to add anything if something breaks 💜

### DIFF
--- a/src/audio/miniaudio/ma_audio_system.c
+++ b/src/audio/miniaudio/ma_audio_system.c
@@ -100,10 +100,24 @@ static void maInit(AudioSystem* audio, DataWin* dataWin, FileSystem* fileSystem)
     arrput(ma->base.audioGroups, dataWin);
     ma->fileSystem = fileSystem;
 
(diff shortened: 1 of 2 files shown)


Un1q32:
Regarding FreeBSD 2.2.8, I don't think audio works on 2.2.8 anyway, it was tested with audio disabled.

Un1q32:
And regarding old Linux, miniaudio doesn't compile on Linux distros older than about 10 years.  Old Linux should use the OpenAL audio backend.

Un1q32:
In general, miniaudio doesn't have very good support for old operating systems, except windows, which seems to work all the way down to NT 3.1.

PR #281, opened 2026-06-29
replace alloca with malloc

Un1q32:
alloca is unsafe, there could be a stack overflow on high instCount and we wouldn't know.
I don't think this is a hot enough path for this to really affect performance.  If it is than I can rework this to actually do what the comment said was being done but wasn't, stack allocating on smaller amounts and heap allocating for large amounts.


Fancy2209:
I trust it

PR #286, opened 2026-07-01
Update readme list of cli options

Un1q32:
This old list sucked.  Basically copies the `--help` usage message into the readme.


MrPowerGamerBR:
I would put the CLI arguments themselves in a `inline code`

PR #288, opened 2026-07-01
modern-gl: use vertex struct instead of array of floats

cobaltgit:
32 bytes -> 20 bytes per vertex, minor memory savings but free nonetheless

### DIFF
--- a/src/gl/gl_renderer.c
+++ b/src/gl/gl_renderer.c
@@ -20,7 +20,6 @@
 
 // ===[ Constants ]===
 #define MAX_QUADS 4096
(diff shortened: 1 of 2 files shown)


Un1q32:
I tested this a little bit, everything seems to render fine and it saves about 0.3 MB of RAM in the undertale opening (the total use is about 264 MB).

cobaltgit:
Ah well I suppose it's a free memory saving either way

cobaltgit:
Also doesn't do as many floating point ops which should be good for older CPUs

PR #289, opened 2026-07-02
gl_wrappers.h

Un1q32:
Moves all those wrapper function things into their own header.
Reduces code duplication in gl_legacy_renderer.c
Also fixes stuff like glBindFramebuffer being called unconditionally in gl_common.c and desktop/main.c.

### DIFF
--- a/src/desktop/main.c
+++ b/src/desktop/main.c
@@ -35,6 +35,7 @@
 #ifdef ENABLE_LEGACY_GL
 #include "gl_legacy_renderer.h"
 #endif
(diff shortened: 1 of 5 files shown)


Un1q32:
At some point, someone should track down drivers that have GL_EXT_framebuffer_object and not GL_ARB_framebuffer_object so we can know for sure if the EXT API is close enough to the ARB API to work as-is.

cobaltgit:
Looks good to me, reckon things like the gma x3100 will have EXT framebuffer objects but not ARB

PR #295, opened 2026-07-03
Apply an offset to certain primitives just like GameMaker runner does

iProgramMC:
Controlled by a flag, because some GameMaker games don't actually need this fix applied.

This fixes the gears in the Core in UNDERTALE and makes them look correct.

### DIFF
--- a/src/runner.c
+++ b/src/runner.c
@@ -2426,10 +2426,7 @@ void Runner_initFirstRoom(Runner* runner) {
 
     // GameMaker Studio (as of 2024.14) applies an offset of 1 virtual pixel to certain
     // primitives if this flag isn't set.
-    runner->applyOffsetForPrimitives = true;
-    if (dataWin->optn.info & 0x800000000LL) {
-        runner->applyOffsetForPrimitives = false;
-    }
+    runner->applyOffsetForPrimitives = !(dataWin->optn.info & 0x800000000LL);
 
     // Run global init scripts with the global scope instance as "self"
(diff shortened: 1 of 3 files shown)


src/runner.c
```diff
@@ -2424,6 +2424,13 @@ void Runner_initFirstRoom(Runner* runner) {
 
     runner->gameStartTime = nowNanos();
 
+    // GameMaker Studio (as of 2024.14) applies an offset of 1 virtual pixel to certain
+    // primitives if this flag isn't set.
+    runner->applyOffsetForPrimitives = true;
```
Un1q32:
```suggestion
    runner->applyOffsetForPrimitives = !(dataWin->optn.info & 0x800000000LL);
```

iProgramMC: done

Un1q32:
LGTM.  Tested in modern-gl and legacy-gl, does indeed fix the gears right before muffet.

Un1q32 (approved):

iProgramMC:
Update: the condition is actually: `if (GameMakerVersion < 2024.14 || (Flags & 0x800000000)) then ApplyFix = true;`

Fancy2209:
Can you add it to this PR? So we merge this and then the fix for GMS version.

iProgramMC:
@Fancy2209 Yup, I've updated the condition, and also fixed an issue with draw_rectangle that I added due to tunnel vision.

PR #298, opened 2026-07-05
stbi config

Un1q32:
defining STBI_NO_STDIO is a very small but totally free improvement in binary size, a little less than 700 bytes on glfw3 linux x86_64.  Comes from excluding code to parse images from FILE*, but we always pass raw memory buffers and never use those functions, so they're useless to us.

The STBI_NEON thing just enables SIMD instructions on armv7 and arm64 automatically.

### DIFF
--- a/src/android/stb_impl.c
+++ b/src/android/stb_impl.c
@@ -1,4 +1,5 @@
 #define STB_IMAGE_IMPLEMENTATION
+#define STBI_NO_STDIO
 #include "stb_image.h"
 
 #define STB_IMAGE_WRITE_IMPLEMENTATION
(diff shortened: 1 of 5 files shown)


cobaltgit:
Afaik stbi_neon only really accelerates JPEG decoding right?

Un1q32:
> Afaik stbi_neon only really accelerates JPEG decoding right?

Idk, we compile in jpeg support anyway so even if its actually just that its still helpful
I looked into disabling jpeg support to free more space, but in my brief research it seems you can use jpeg textures in gamemaker with the sprite_add function, which is currently stubbed but we might add in the future.

cobaltgit:
Yea sounds fair, good to have accelerated jpeg decoding either way

PR #299, opened 2026-07-05
memory usage info on backspace + free blobData after we're done with it

Un1q32:
At some point we should add a proper debug overlay to desktop/modern-gl, and we can put memory usage info in that
memory usage only works on Linux for now


Un1q32:
Freeing blobData saves a lot of ram, amount depends on the game, but in undertale it freed 12.1 meg.

cobaltgit (approved):
Free RAM is free RAM..green light

PR #305, opened 2026-07-05
disable chapter4 test

Un1q32:
I've run it on my own machine and it doesn't crash, I have no idea why it's crashing on the test thingy

### DIFF
--- a/tests/tests.conf
+++ b/tests/tests.conf
@@ -94,21 +94,21 @@ tests = [
             ]
         ]
     },
-    {
-        name = "Test if path_add is creating paths closed by default (DELTARUNE Chapter 4 - Jackenstein)"
-        commercialGame = true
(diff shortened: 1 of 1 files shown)


Fancy2209:
We should try to figure out why it crashes instead of just disabling if and forgetting about it then...

MrPowerGamerBR:
It crashes because the game calls `audio_set_master_gain` while the tests are ran with `--headless`, which use the `NoopAudioSystem`, and because of that it segfaults trying to call a nullptr

cobaltgit:
Lmao I forgot to implement those

cobaltgit:
Myyyy bad

Un1q32:
> Lmao I forgot to implement those

do it pls

cobaltgit:
done the thing

Un1q32:
tests still failing bruh

PR #306, opened 2026-07-06
remove useless calloc

Un1q32:
Saves 3 256 byte allocations, and having to zero those allocations, and having to free those allocations, per frame


cobaltgit (approved):
Green light

The mind boggles at heap allocating a small compile-time constant every frame

PR #307, opened 2026-07-06
iOS port

Un1q32:
horrifically vibecoded
to add a game put the game folder in /var/mobile/Documents/Butterscotch/games, or <the app's documents folder>/games.  You can also use the files app to add games on iOS 11+.

TODO:
- actually understand all the code, theres a good chunk right now that AI wrote and I never understood.


stossy11:
I would say take a look at MobileGLuES + MetalANGLE for rendering, that had fixed some graphical issues for me on my iOS port (OpenGLES is deprecated on most modern iOS versions), especially with the start of DELTARUNE: Chapter 3 and had allowed the use of the Metal HUD for Performance Information.

Un1q32:
> I would say take a look at MobileGLuES + MetalANGLE for rendering, that had fixed some graphical issues for me on my iOS port (OpenGLES is deprecated on most modern iOS versions), especially with the start of DELTARUNE: Chapter 3 and had allowed the use of the Metal HUD for Performance Information.

One of my goals for this port is to support every iOS device, including the original iPhone through software rendering, as well as to support iOS 2.  Additionally, while not quite as important, I would like to support building with the iOS 2.0 SDK, and maybe even with Apple's original GCC 4.0 iOS compiler.  Something tells me that those libraries wouldn't be happy building under those constraints.  If you know, could you tell me what iOS versions each library supports, it might be worth looking into as an optional dependency even if the minimum version is high.  I haven't noticed any graphical issues with the modern-gl renderer, but I also haven't played with it that much.  What's broken in chapter 3?

stossy11:
> > I would say take a look at MobileGLuES + MetalANGLE for rendering, that had fixed some graphical issues for me on my iOS port (OpenGLES is deprecated on most modern iOS versions), especially with the start of DELTARUNE: Chapter 3 and had allowed the use of the Metal HUD for Performance Information.
> 
> One of my goals for this port is to support every iOS device, including the original iPhone through software rendering, as well as to support iOS 2. Additionally, while not quite as important, I would like to support building with the iOS 2.0 SDK, and maybe even with Apple's original GCC 4.0 iOS compiler. Something tells me that those libraries wouldn't be happy building under those constraints. If you know, could you tell me what iOS versions each library supports, it might be worth looking into as an optional dependency even if the minimum version is high. I haven't noticed any graphical issues with the modern-gl renderer, but I also haven't played with it that much. What's broken in chapter 3?

from my testing, atleast on modern iOS, chapter 3 just doesn't render at all past the DELTARUNE text screen. MobileGLuES + MetalANGLE only supports iOS 14+ afaik, so you could just omit it for any version below.

Un1q32:
> from my testing, atleast on modern iOS, chapter 3 just doesn't render at all past the DELTARUNE text screen. OpenGLuES + MetalANGLE only supports iOS 14+ afaik, so you could just omit it for any version below.

Chapter 3 has worked and rendered correctly on every iOS version I've run it on, 10, 12, 14, 18, 26, as well as on an apple silicon mac.
Maybe try rebasing your iOS port on upstream, the modern-gl renderer has gone through some changes.

stossy11:
> > from my testing, atleast on modern iOS, chapter 3 just doesn't render at all past the DELTARUNE text screen. OpenGLuES + MetalANGLE only supports iOS 14+ afaik, so you could just omit it for any version below.
> 
> Chapter 3 has worked and rendered correctly on every iOS version I've run it on, 10, 12, 14, 18, 26, as well as on an apple silicon mac. Maybe try rebasing your iOS port on upstream, the modern-gl renderer has gone through some changes.

yeah it could've probably been because my port is a bit old now. i haven't touched it in a while.
last i touched it was to make this post https://x.com/Stossy12/status/2053600784789688637 back in may

Un1q32:
Oh wait are you talking about the first tenna cutscene thing, yeah that doesnt work even on macOS with GLFW3.

Un1q32:
No, thats missing the video_open functions, it doesn't work on any OS or renderer

stossy11:
> Oh wait are you talking about the first tenna cutscene thing, yeah that doesnt work even on macOS with GLFW3.

nah i was talking about the start of the chapter, before we even meet tenna.

Un1q32:
By the way, if you do rebase your iOS port could you show off how well it performs on lower end devices when using Apple's GLES driver?  My iOS port seems to perform over 10x worse than desktop OpenGL on my M4 MacBook, and my most powerful iOS devices can't quite run the chapter 5 festival area at full speed.  I'm pretty sure it's just Apple's GLES driver being bad but it would be nice to eliminate the possibility that I'm doing something stupid in the iOS specific code.

Un1q32:
Just to be clear, this isn't ever intended to be merged.  It's left here for anyone who really wants an iOS port that's somewhat up to date, and for me to be able to test how well butterscotch performs on iOS.  Eventually there will hopefully be an actually good iOS port, and when that happens this will be abandoned.

PR #315, opened 2026-07-11
Allocate room stuff in 1 block to reduce heap fragmentation

Un1q32:
I can't tell exactly how much memory this saves, but it definitely saves something.


Un1q32:
This appears to save ~400k when running undertale with lazy rooms off, which seems odd because that would mean it's saving more than 1k per room, but whatever.

Un1q32:
Realistically, the absolute *minimum* this can save is 16 bytes per room on a 32 bit system with a magic malloc implementation that never fragments and has an allocation overhead of only 1 size_t.  In undertale that would be about 5k, and it's free, so whatever.

cobaltgit:
Looks good to me, not sure of the impact with the new mmap stuff but go ahead

PR #316, opened 2026-07-13
mmap data.win and use it for audo

Un1q32:
Saves a LOT of memory on systems that can do it.


cobaltgit (approved):
So far so good, definitely gonna test this on my 128MB RAM handhelds as soon as this is merged in

PR #318, opened 2026-07-15
refactor(android): use Glad instead of hard-coded GLES3

cobaltgit:
Android should now support GLES 2.0


Un1q32:
This is prolly good but I can't test RN.

Un1q32:
I think we should just leave this backend alone until we have a new frontend for android.  Then we either make the new frontend use this backend or make a new backend and kill the old one.

cobaltgit:
> This is prolly good but I can't test RN.

I did test it working on my OnePlus 13 when I first made this branch, should work on any GLES3 device in theory

PR #321, opened 2026-07-16
Log system instead of printf and fprintf(stderr, ...)

Purplesmaug96:
Only tested it on desktop/sdl3, seems to work fine there
Logs to ./butterscotch.log by default


Un1q32:
It should log to stdout or stderr by default still.  I do like the idea of dedicated logging functions though.  Would be useful to for example make iOS print everything to the syslog.

Purplesmaug96:
it should do, stdout for normal logs and stderr for warnings and errors

Purplesmaug96:
there's also purple debug log functions now

Un1q32:
Colors are nice, they should be disabled if the output file isnt a TTY.

Un1q32:
I will test this soon

Purplesmaug96:
colours are disabled by default for files, cli argument to control it

src/ps3/main.c
```diff
@@ -310,7 +311,7 @@ int main(int argc, char* argv[]) {
         bool shouldStep = true;
         if (runner->debugMode && debugPaused) {
             shouldStep = RunnerKeyboard_checkPressed(runner->keyboard, 'O');
-            if (shouldStep) fprintf(stderr, "Debug: Frame advance (frame %d)\n", runner->frameCount);
+            if (shouldStep) Log_logDebug("Debug: Frame advance (frame %d)\n", runner->frameCount);
```
Un1q32: just have the logDebug function insert "Debug: " into the start of the message itself so we don't have to do it every time.

Also, could you name the functions something like
logInfo
logWarn
logError
logDebug

Purplesmaug96: should things like Runner_dumpState and DataWin_printDebugSummary use logDebug with the "Debug: " prefix or just logInfo

Un1q32: logInfo

Un1q32: logWarn should insert "Warning: " before the messages simalarly to how logDebug does, and the "Warning" text should be yellow if the output is a TTY.  logError is the same but with "Error" and red.

Purplesmaug96: should be done

Un1q32:
For detecting if output is a terminal, use the isatty function from the C library.

Un1q32:
Ok I see now this is totally different from what I had in mind.

There should be a function called platformLog that handles all logging, defined by the backend.  On the normal desktop backends that should be
```c
void platformLog(, const char *format, ...) {
    va_list ap;
    va_start(ap, format);
    vfprintf(stderr, format, ap);
    va_end(ap);
}
```

All the log* functions should go through that platformLog to output text.

On other backends that can be something that calls syslog instead, or the android logging functions, or outputs to a file.
We should *not* be logging to stdout *and* a file by default.
If you need to output to a file with the desktop backend then you should redirect stderr in your shell.

Purplesmaug96:
is it ok if i make it "void platformLog(FILE* file, const char *format, ...)" so logInfo can go through stdout since it's not an error or warning

Un1q32:
I would rather you replace the FILE* with an enum for the error type, since FILE* only makes sense in the printf context.

Un1q32:
Remove the file log stuff.  If you wanna log to a file you can do `./butterscotch > out.log 2>&1` or `./butterscotch 2>&1 | tee out.log`.

src/log.h
```diff
@@ -0,0 +1,58 @@
...
+#define LOG_MAX_FILES 16
+
+typedef enum {
+	LOG_TYPE_NORMAL=0,
+	LOG_TYPE_WARNING=1,
+	LOG_TYPE_ERROR=2,
+	LOG_TYPE_DEBUG=3
+} logType;
+
+typedef enum {
```
Un1q32: remove this

src/log.c
```diff
@@ -0,0 +1,282 @@
...
+#define va_copy(d, s) ((d) = (s))
+#endif
+
+#define ANSI_COLOUR_CODE_WHITE "\x1b[0;37m"
+#define ANSI_COLOUR_CODE_BOLD_YELLOW "\x1b[1;33m"
+#define ANSI_COLOUR_CODE_BOLD_RED "\x1b[1;31m"
+#define ANSI_COLOUR_CODE_BOLD_PURPLE "\x1b[1;35m"
+
+// In the platform main.c
+void platformLog(const logType type, const logOutType out, const char *format, va_list va);
```
Un1q32: remove the logOutType argument

src/desktop/main.c
```diff
@@ -302,6 +309,26 @@ static void printOsTypeNames(FILE* out) {
     }
 }
 
+#ifndef va_copy
+#define va_copy(d, s) ((d) = (s))
+#endif
+
+void platformLog(const logType type, const logOutType out, const char *format, va_list va) {
```
Un1q32: replace with
```c
void platformLog(const logType type, const char *format, va_list va) {
	vfprintf(type == LOG_TYPE_NORMAL ? stdout : stderr, format, va);
}
```

Purplesmaug96:
can i please use dup2 or something to redirect stdout/err to a file

Un1q32:
What use case is there?  Maybe for other backends where stdout is less accessible like mobile or consoles it would make sense, but for the desktop backend? why?

Un1q32:
Oh wait, is it windows? its windows isnt it.

Purplesmaug96:
mainly just want to be able to view it in my ide without "./butterscotch 2>&1 | tee out.log" 
Edit: no i use arch (btw) but if i ever am forced to use butterscotch on windows it would be useful, although you can get it to display a second terminal window for stdout

Un1q32:
What fucked up and evil IDE do you have that won't show you stdout?!??!

Un1q32:
Doesn't every IDE have an integrated terminal or at least launches a terminal along with the app?

Purplesmaug96:
i use vscodium and just launch my stuff on konsole normally

Un1q32:
I used to use vscode it has an integrated terminal, why not just use that?

Purplesmaug96:
too small, if i make it bigger it takes up the code space
not on a second monitor
i think its sandboxed so wont work with msvc via wine

Un1q32:
Can't you configure vscode to launch an external terminal?

Purplesmaug96:
probably but i dont wanna it works fine as is

Un1q32:
Just use tee, setup an alias in your shell to make it easier to type.  `alias bscotch='./butterscotch 2>&1 | tee out.log'`

Purplesmaug96:
nah I'll just deal with there not being a log file or add a temporary write to file shim thing

src/log.h
```diff
@@ -0,0 +1,31 @@
+#ifndef _BS_LOG_H
+#define _BS_LOG_H
+
+#include <stdio.h>
+#include <stdbool.h>
+#include <stdarg.h>
+
+#define LOG_MAX_FILES 16
```
Un1q32: unused?

Un1q32:
Since you need to check if stdout is a tty for weather color should be enabled, and because color might not be desirable on some platforms, you should make color get handled in platformLog instead of in the dedicated log functions.

Un1q32:
Here's a better platformLog implementation actually
```c
void platformLog(const logType type, const char *format, va_list va) {
    FILE *out = stderr;
    switch (type) {
        case LOG_TYPE_NORMAL:
            out = stdout;
            break;
        case LOG_TYPE_WARNING:
            fputs("\033[33mWarning:\033[0m ", out);
            break;
        case LOG_TYPE_ERROR:
            fputs("\033[31mError:\033[0m ", out);
            break;
    }
    vfprintf(out, format, va);
}
```
This also properly resets the color which wasn't being done before.

src/log.h
Purplesmaug96: yeah forgot to remove it

Un1q32:
This appears done waiting on approval from others

cobaltgit (approved):
Looks good to me, tested on glfw3 and sdl3 so far

PR #322, opened 2026-07-16
vm: only check for rem by zero for integers, and maybe a couple other chapter 5-related bits

cobaltgit:
Fixes Deltarune Ch5 crashing when entering the garden of hopes and dreams...but

image


Fancy2209:
Put the new tile functions on a separate PR please.

Ananim353:
The handleRem one looks wrong though: `Rem` is integer division (`div`), not the remainder.
Built the project once with Igor, so the same data.win runs under the GameMaker runtime
(com.gamemaker.exe) and under Butterscotch, and diffed the output line by line:

    7 div 2 = 3    -7 div 2 = -3    7.5 div 2 = 3
    7 mod 2 = 1    -7 mod 2 = -1    7.5 mod 2 = 1.5    7 % 2 = 1

Butterscotch matches on all seven today. With fmod in handleRem, `7 div 2` would return 1.

Fancy2209:
What's the current status of this

emiyl:
the rem zero bypass isn't needed anymore as i fixed the underlying issue

PR #326, opened 2026-07-17
feat(audio): allow lazy loading of AUDO chunk

cobaltgit:
largely based on what [that one Vita port](https://github.com/WolffsRoom/DeltaruneVita/blob/main/src/butterscotch) is doing with the OpenAL backend, useful for platforms like the Vita that can't mmap the way POSIX and Win32 do

Also mmapping the audio groups (`audiogroup*.dat`) slightly saves memory


Un1q32:
Where does the mmap audio groups thing happen?  I can't see it.  Am I a stupid illiterate chud who can't read code?

cobaltgit:
The audio groups are data.wins here so we just pass the MAP_FILE mode to the parse options

cobaltgit:
Well not sure what OpenAL does now so

Un1q32:
Could you seperate the mmap audio groups thing into its own patch?  I wanna see how much memory that saves on it's own.

cobaltgit:
#352

src/desktop/main.c
```diff
@@ -420,6 +421,7 @@ static void printUsage(const char *argv0) {
         "    --save-folder <directory>              - Set the directory will save files will be stored\n"
         "    --game-args <args>                     - Arguments to pass to the game\n"
         "    --lazy-textures                        - Load textures into VRAM on first use, improving startup times\n"
+        "    --lazy-audio                           - Load audio into RAM on first use, reducing memory usage\n"
```
Un1q32: trailing whitespace

Un1q32 (commented):
LGTM, fancy should have a look tho.

PR #328, opened 2026-07-18
update miniaudio to 22b9d84622adf6f00350eff3845784c8c4c79f6e

cobaltgit:
now with psvita support


Un1q32:
You'll have to add back our patches for MSVC and CRTDLL support

vendor/miniaudio/miniaudio.h
```diff
@@ -44361,7 +44361,7 @@ static void ma_device_prime_playback_buffer__oss(ma_device* pDevice)
                 break;
             }
     
-            ma_int32 rowsWritten = write(pDeviceStateOSS->fdPlayback, buffer, framesToWriteThisIteration * bpf);
+            ssize_t rowsWritten = write(pDeviceStateOSS->fdPlayback, buffer, framesToWriteThisIteration * bpf);
```
Un1q32: this type doesn't exist on msvc btw, so if this isn't in unix-only code it should be changed to `long` or something.

Un1q32: Oh well its calling `write` so it has to be unix-only, sorry

Un1q32: you can make it ssize_t, its better that way

Un1q32:
You need to add the `__CRTDLL__` thing back

Un1q32:
This got done by the vita port PR

PR #329, opened 2026-07-19
feat(desktop): allow launching in fullscreen

cobaltgit:
Useful for linux handhelds that don't use a window manager. Tested working on SDL 1.2, SDL2, SDL3 and GLFW3

Fixes the below, pictured is Undertale (640x480) running on a TrimUI Brick Pro (1024x768) via the SDL2 desktop backend without the `--fullscreen` flag.
image

### DIFF
--- a/src/desktop/backends/glfw2.c
+++ b/src/desktop/backends/glfw2.c
@@ -91,6 +91,8 @@ void platformGetMousePos(double *xPos, double *yPos) {
     *yPos = (double)my;
 }
 
+static bool glfw2Fullscreen = false;
+
 static bool platformGetWindowFocus(void) {
     return glfwGetWindowParam(GLFW_ACTIVE);
 }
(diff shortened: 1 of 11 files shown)


Un1q32:
Can you implement the is fullscreen check function in gamemaker, and hook this up to it.
Also make sure this sets the correct resolution on HiDPI stuff.
Also, should we just fucking give up and disable HiDPI support? It makes the window way too small to play with on macOS, and makes stuff like this a nightmare to develop.  There was a PR forever ago that tried to do all the fullscreen functions, and it couldn't get finished because of issues with HiDPI.

Fancy2209:
What happens on PS2 and PS3?

cobaltgit:
They're stubbed out (checks if the fullscreen runner functions set in the desktop platform are nullpointer and does nothing if they are)

PR #330, opened 2026-07-19
perf(gl): improved state tracking and other optimisations

cobaltgit:
Todo: apply to legacy-gl and test

### DIFF
--- a/src/gl/gl_renderer.c
+++ b/src/gl/gl_renderer.c
@@ -73,6 +73,26 @@ static inline uint8_t floatToUnormByte(float v) {
     return (uint8_t)(v * 255.0f + 0.5f);
 }
 
+// ===[ State Tracking Wrappers ]===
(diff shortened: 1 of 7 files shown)


cobaltgit:
Slight but free performance boost here

cobaltgit:
There's a lot more glEnable calls in legacy-gl, could maybe do a bitmask instead of however many booleans

EDIT: No point, it's not a hot struct and you'd only be saving 3 bytes and having to map the capabilities from GL to the bitmask.

cobaltgit:
Not enough meaningful improvement yet

cobaltgit:
Will look back into this later

cobaltgit:
Ch5 Hometown performance, should be a similar minor increase on paper in scenes with active shaders

| **Before** | **After** |
|------------|------------|
| image | image |

PR #334, opened 2026-07-22
Implement string_ends_with

basiccube:


src/vm_builtins.c
```diff
@@ -2255,6 +2255,23 @@ static RValue builtin_string_starts_with(MAYBE_UNUSED VMContext* ctx, RValue* ar
...
+// Source - https://stackoverflow.com/a/744822
+static RValue builtin_string_ends_with(MAYBE_UNUSED VMContext* ctx, RValue* args, int32_t argCount) {
+    if (2 > argCount) return RValue_makeInt32(0);
+    char* str = RValue_toString(args[0]);
+	char* substr = RValue_toString(args[1]);
+
+	size_t strLen = strlen(str);
+	size_t substrLen = strlen(substr);
+	if (substrLen > strLen)
+		return RValue_makeBool(false);
```
cobaltgit, line 2269: You might wanna either free the (sub)str here before early returning or just initialise ret beforehand and set it to false here

```suggestion
                free(substr);
                free(str);
                return RValue_makeBool(false);
```

cobaltgit (changes requested):

src/vm_builtins.c
```diff
@@ -2255,6 +2255,23 @@ static RValue builtin_string_starts_with(MAYBE_UNUSED VMContext* ctx, RValue* ar
...
+    bool ret = (strncmp(str + strLen - substrLen, substr, substrLen) == 0);
```
Un1q32: You should use memcmp instead of strncmp, memcmp should be faster.

src/vm_builtins.c
```diff
@@ -2255,6 +2255,26 @@ static RValue builtin_string_starts_with(MAYBE_UNUSED VMContext* ctx, RValue* ar
     return RValue_makeBool(ret);
 }
 
+// Source - https://stackoverflow.com/a/744822
+static RValue builtin_string_ends_with(MAYBE_UNUSED VMContext* ctx, RValue* args, int32_t argCount) {
+    if (2 > argCount) return RValue_makeInt32(0);
```
Un1q32: Is returning an Int32 here intentional?

basiccube: it was copied over from string_starts_with, so I'm not sure, probably not intentional?

PR #335, opened 2026-07-23
fix(vm): guard room builtins before current room init

emiyl:
This fixes a crash during global init where scripts try to write to room builtins before currentRoom has been initialised. It tries to resolve a safe target for pre-room access and defaults to the FPS value when not available.

In my testing, this issue would trigger when using the 60FPS mod.


Fancy2209 (approved):
Sounds like we should have always done this

cobaltgit (approved):

PR #338, opened 2026-07-23
string_starts_with fix

basiccube:
Arguments are backwards, should be (str, substr)
Also changed the returned value when there are less than 2 arguments to be a bool rather than int32 + replaced strncmp with memcmp


Fancy2209:
what's the point of replacing strncmp with memcmp?

Un1q32:
> what's the point of replacing strncmp with memcmp?

speed, memcmp does less

Fancy2209 (approved):
Oh I see, LGTM then

PR #339, opened 2026-07-24
fix(layer): make layer_destroy work for room layers

emiyl:
Fixes collision in Chapter 5 hometown during the festival. The room controller calls `layer_destroy("OBJECTS_MORNING_COLLISION")` but it wasn't doing anything before. Now it does!

https://github.com/user-attachments/assets/6f8e0100-319d-4c07-accd-b839f26a04e6


Fancy2209 (approved):
Looks good to me, thank you so much.

PR #340, opened 2026-07-24
fix(vm): correct BC17 DUP swap-mode operand decoding

emiyl:
Fixes crash in chapter 5 when building Susie's outfit

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -57,6 +57,20 @@ static int gmlTypeNativeSize(uint8_t gmlType) {
     }
 }
 
+static const char* gmlTypeName(uint8_t gmlType) {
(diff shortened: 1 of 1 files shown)


Fancy2209:
Not super well versed in the VM, will research this a bit to be sure before merging!

Un1q32:
You'll need to fix CI btw

emiyl:
Marked as draft as it seems to break erasing saves

cobaltgit:
Well it's the save menu itself I think, I haven't tested trying to select an option other than Erase but I reckon it'll be the same

emiyl:
Closing this as better solutions seem to have been found on Discord

PR #343, opened 2026-07-24
fix(paths): Set default speed of PathPositionResult to 100

emiyl:
Default value of speed on a point is 100, not 0.

Tested on GameMaker Runtime 2026.0.0.23, using the following script:

```gml
var p = path_add();

path_add_point(p, 0, 0, 250);
path_clear_points(p);

show_debug_message(path_get_speed(p, 0));
```

In GameMaker, this outputs 100. In Butterscotch, this would previously output 0.


Un1q32:
GCC 2.95 doesn't like your struct creation syntax (its C99)
do
```c
PathPositionResult result = {0};
result.speed = 100.0f;
```

PR #352, opened 2026-07-24
perf(audio): mmap audiogroup*.dat if main data.win is being mapped

cobaltgit:


Un1q32:
seems fine but ill let fancy have the final say because he knows more about the audio backends

Un1q32:
Saves like 8 meg in chapter 3 btw

PR #363, opened 2026-07-28
Added RLE tile data checks if it's GM 2024.2 or 2024.4 if aligned

AloXado320:
Fixes games made in a lower 2024.x version being misreported as 2023.2, causing garbled tile data being rendered.

Checks are taken from UTMT since the version was being reported correctly there.

Tested on MotionRec Demo, where the background wasn't showing properly.

### DIFF
--- a/src/data_win.c
+++ b/src/data_win.c
@@ -2102,7 +2102,6 @@ static void parseROOM(BinaryReader* reader, DataWin* dw, bool lazyLoadRooms, Str
             BinaryReader_skip(reader, 22 * 4);
 
             uint32_t layersPtr = BinaryReader_readUint32(reader);
-            uint32_t seqnPtr = BinaryReader_readUint32(reader);
 
             BinaryReader_seek(reader, layersPtr);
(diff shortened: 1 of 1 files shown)


Fancy2209:
We're MPL2 now 
Is this allowed?

cobaltgit:
MPL 2.0 is compatible with GPL 3.0 sooo

Un1q32:
> MPL 2.0 is compatible with GPL 3.0 sooo

Yes but the GPLv3's virality would mean the whole project is GPLv3 until we remove or rewrite this code.

Fancy2209:
If Butterscotch Android on Google Play is dead should we revert to GPLv3

Un1q32:
We can, it would lock us out of linking to any GPLv2 libraries in the future, but other than that I see no reason not to.

cobaltgit:
> If Butterscotch Android on Google Play is dead should we revert to GPLv3

No reason not to, our next big library link will end up being ffmpeg which at worst is GPL-2.0-or-later (without encoders, which is our use case, it'll be LGPL-2.1-or-later)

PR #365, opened 2026-07-28
Implement move_and_collide and cleanup related collision functions

AloXado320:
Fixes movement on MotionRec Demo, at least on the first screen stage, since going past it collision doesn't work properly.

Also cleanup related collision functions that use similar checks using `performCollisionTest`, like in HTML5.


AloXado320:
Collision doesn't work properly on that game due to tilemap collisions being unimplemented, so this changes would be mostly cleanup then I'll leave tilemap collision implementation for another PR

PR #367, opened 2026-07-29
fix(collision): sort events in ascending order

cobaltgit:
fixes my own GM2026 game that uses inheritance for projectiles: before, rock -> projectile collision fired but mine (child of projectile) -> rock did not, thanks deepseek.


cobaltgit:
Worst case scenario if this isn't the case I screw off inheritance and do it the old way.

cobaltgit:
Seems to pass the Undertale long corridor column test

PR #369, opened 2026-07-30
PlayStation Vita port

hatoving:
Here's a barebones and a bit untested port of this project for the PS Vita.

It uses VitaGL to handle the graphics. I've added "support" for OpenAL and Miniaudio-- support in quotations because while OpenAL does work, I couldn't get Miniaudio to do the same. Gamepad should work as expected. I stole some code from the other backends (mainly the Desktop and PS2 backends) and rewrote them for the Vita (you can probably see what I mean when you start noticing the fact that I forgot to remove some comments...), like for example the gamepad code.

I've also implemented `game_change`-- it works, although for games like DELTARUNE it can't quite load in music files from another directory due to the game trying to check something via a struct that Butterscotch hasn't implemented in yet...? I could be wrong on that or I'm simply misunderstanding it.

The only issues I've encountered are some games not loading due to either Butterscotch not supporting them yet or because the Vita doesn't have sufficient memory, although some of the lazy loading options did help sometimes (particularly, lazily loading in the textures for Pizza Tower helped it quite a bit... takes forever to load though...)
Another current issue really is that if the game resizes its screen, then it'll look really weird on the Vita. So long it doesn't try to do that, it looks just fine...

Please keep in mind that I'm not the best at C, so if you see some absolutely horrendous code, then I'd love to see it get cleaned and fixed up by someone who knows any better lol.

Right now, if you want to test it out yourself, simply put game files in `ux0:data/butterscotch/`. This'll act as the save directory, too. In the end your `data.win` file should be in that folder. I've attached a ZIP file that has a VPK file in it for those who just want to test it out right away.

If you want to build it, use this: `cmake -DPLATFORM=vita -DAUDIO_BACKEND=openal -DCMAKE_BUILD_TYPE=Debug ..`
For the Vita specifically you HAVE to specify an audio backend, otherwise it'll choose none. As mentioned before, I've only gotten OpenAL to work.

Here are some photos of the games I've tested:
https://cdn.discordapp.com/attachments/712548851248005140/1532241945982668991/IMG_20260730_062133.jpg?ex=6a6c2303&is=6a6ad183&hm=b3b644bbd2ca71dd9e7ccd332c4fd8e59443cd70ba991b08776661d2212d675c&
https://cdn.discordapp.com/attachments/712548851248005140/1532241946477723668/IMG_20260730_062128.jpg?ex=6a6c2303&is=6a6ad183&hm=ff35b330faee164a679f5d13f8b09d5e2e232fb3727f7cb794fd40da2779384b&
https://cdn.discordapp.com/attachments/712548851248005140/1532241947094159380/IMG_20260730_062050.jpg?ex=6a6c2303&is=6a6ad183&hm=f7a23b41743ddfe035623a2f65d7a01003fb46bea4be764098d91d17825ecdce&

I'd love to help on this further if I did a good enough job! This project is great and it allows for some really awesome stuff.

[Butterscotch.zip](https://github.com/user-attachments/files/30531458/Butterscotch.zip)


hatoving:
Update! Miniaudio works now thanks to @cobaltgit's PR!
https://github.com/ButterscotchRunner/Butterscotch/pull/328

hatoving:
Update, recommended way to build this would be with:
``cmake -DPLATFORM=vita -DAUDIO_BACKEND=miniaudio -DCMAKE_BUILD_TYPE=Debug ..``
vitaGL should be built with:
``make clean && make NO_DEBUG=1 NO_SPLASHSCREEN=1 LOG_ERRORS=1``

I've also added preprocessor support. Right now you can generate a "textures.bin" file from a data.win and just put it in the folder as that file, and my port will automatically use that instead for any texture loading. Supports PVR files and PNG files as well, just in case.

PR #370, opened 2026-07-30
Stack args

Purplesmaug96:
Stack alloc args in handleCall and handleCallV if there are less than VM_MAX_STACK_ARGS arguments
VM_MAX_STACK_ARGS is #define-d in vm.h as 8


cobaltgit:
Looks good to me, reckon this could also be applied to the script args and local vars again

Un1q32:
Is there any easy way you could make extra args overflow into the heap allocation instead of only using the heap allocation, might save a bit of memory, if it adds a lot of complexity then prolly not worth it tho.

Un1q32 (approved):
lgtm

PR #371, opened 2026-07-30
Skip drawing offscreen tiles

Purplesmaug96:
Improves performance for me in modern gl by a bit.

Done by @Crazywastaken, they couldnt figure out how to make a pr so im doing it

### DIFF
--- a/src/runner.c
+++ b/src/runner.c
@@ -59,7 +59,7 @@ void Runner_updateCameraViewSimple(GMLCamera* camera) {
 
     Matrix4f projectionMatrix;
     Matrix4f_Orthographic(&projectionMatrix, (float) camera->viewWidth, -((float) camera->viewHeight), 32000.0, 0.0);
-    
+
 
     camera->viewMatrix = viewMatrix;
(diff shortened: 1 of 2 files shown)


Crazywastaken:
hi

Crazywastaken:
i could've made a branch then a pr, but ok

Purplesmaug96:
you can still do that

Crazywastaken:
> you can still do that

yeah but its too late at this point, its fine (hi discord)

Un1q32:
this makes no performance difference in my testing on WSL2 with modern-gl, will test more stuff later

Purplesmaug96:
apparently breaks stuff

MrPowerGamerBR:
It would be better to just cull tiles on each renderer, in fact some renderers (like the PlayStation 2 renderer) already cull tiles

Crazywastaken:
> It would be better to just cull tiles on each renderer, in fact some renderers (like the PlayStation 2 renderer) already cull tiles

true

PR #372, opened 2026-07-30
perf(vm): allocate up to 16 script arguments and 64 local variables o…

cobaltgit:
…n the stack before falling back to heap

takes cues from Purplesmaug's stack allocation of function args. Very early builds did this but hard-capped the local variables per code entry to 128 without any heap allocation fallback after that threshold, hamstringing compatibility.

Should help in cases where malloc slogs


cobaltgit:
Was originally done to improve performance on the PS2 target, should apply on similarly low-end targets where the interpreter bottlenecks.

cobaltgit:
Seems to go best with WAD 15/16 games as `resolveLocalSlot` will change the code locals to be heap-allocated atm

PR #376, opened 2026-08-02
[vita] Fix texture preprocessing on legacy GL renderer

hatoving:
I forgot to add some code the legacy GL renderer that loads in the preprocessed textures if they're active. My bad... 🥀


Un1q32:
bleehhhhh

PR #377, opened 2026-08-02
chore(vita): add CI for PS Vita build!

cobaltgit:


hatoving:
just wanted to mention somethin: you have to tell cmake what audio backend it should use for the vita, otherwise itll use none of them. installing and building vitaGL with the commands i mentioned in my PR should prooobabblyyy be on there too 🥴

cobaltgit:
Ok now we're linking with a speedhacked vitaGL - this'll need testing, makes me wish i had a real vita more to test on..that way we can figure out what speedhacks work and what don't, and what could be added to bolster perf

PR #378, opened 2026-08-04
Implement the GameMaker particle system

Ananim353:
`part_*` was not implemented, so any game that uses particles drew nothing where
an effect belonged, with no diagnostic of any kind. DELTARUNE chapters 4 and 5
lose their falling leaves, drifting dust, confetti and cherry blossom petals to
this; Pizza Tower loses its effects the same way.

https://github.com/user-attachments/assets/0ba2a2e0-871b-4e52-82fd-4873f949c8bb

*Chapter 5's garden, running on this branch. Two systems at once: the ambient
`pt_petal` and the whirling `pt_petal_whirl`, both with gravity at an angle on
top of speed wiggle.*

## What this adds

`src/particles.{c,h}` with the three GameMaker resources — systems own emitters
and live particles, types are global so any system can stream any type — plus
the builtins for all three. Both pools live on `Runner` and reuse destroyed ids,
the same tombstone convention the `ds_*` pools already use, because games depend
on it.

Systems with automatic drawing join the depth-sorted drawable list as a new
`DRAWABLE_PARTICLE_SYSTEM` entry, so `part_system_depth` places them among
instances, tiles and layers rather than needing a separate pass. Automatic
updates run once at the end of `Runner_step`, after End Step and before the draw
pass.

Three commits, separable: the core, the rest of the API, and the fixes from a
review pass over the first two.

## Choices worth flagging

**Particles draw from a private random stream, not `rand()`.** Sharing the
global stream would mean every particle spawn shifts the sequence the game
itself sees — so merely adding an effect to a scene would perturb unrelated
randomised behaviour, and every seeded screenshot test with it. The cost is that
`--seed` and `randomize()` do not reach particles.

**A per-system cap of 8192 particles**, warned about once when hit. GameMaker
has no equivalent, but an emitter left streaming in a room the player never
leaves grows without bound, which the console targets cannot absorb.

**Ranges are stored exactly as the game passes them**, without normalising min
against max. GameMaker evaluates `min + random * (max - min)`, and Chapter 4
passes `part_type_direction(-45, -90)`, which sweeps downward; swapping the
bounds would flip the spray.

## Not implemented

`part_type_shape` — it needs the shape textures baked into GameMaker's own
runtime, which cannot be reproduced here. Gaussian and inverse-gaussian emitter
distributions fall back to linear. The GMS 2.3 emitter delay/interval calls, and
`part_type_colour_mix/rgb/hsv`, are absent.

Two known behaviours, both shared with existing drawables rather than introduced
here: two systems at the same depth are ordered by pool slot, which id reuse
makes history-dependent; and changing a system's depth from inside a Draw event
can draw it twice that frame, the same way changing an instance's depth
mid-draw already can.

## Verification

The existing screenshot tests are unaffected — `loritta-and-the-stars` at frame
9000, and the `deltarune-chapter4` jackenstein path test at frames 9473 and
9474, all still match their expected images byte for byte.

Every `part_*` call in DELTARUNE chapters 4 and 5 and in Pizza Tower now
resolves, checked with `--print-unknown-functions`; `vertex_*` still reports as
unknown in the same output, so the check is measuring something.

Pizza Tower is worth calling out because it exercises what DELTARUNE does not:
it never calls `part_system_drawit` or `part_system_update`, relying entirely on
automatic update and automatic draw through the depth list, and it uses
`part_particles_clear` and `part_emitter_destroy`, which DELTARUNE never
touches. Between the two games both spawn paths and both draw paths are covered.

Cross-compiles clean for 32-bit MIPS with `-Wall -Wextra`; the only warnings
attributable to the new file are stb_ds macro expansions that existing files
produce identically.

## One caveat on CI

A pull request from a fork skips the commercial-game downloads, so the two
screenshot tests above will not run here. They pass locally.

---


Un1q32:
Can you restructure this so instead of having particles.c and wrappers in vm_builtins.c, everything is just in vm_builtins.c

Un1q32:
I've heard from emma and casriel that this is good, if there's no opposition in the next couple days or so I'll merge.

AloXado320:
I don't even know how do I locked and limited comments bruh I thought those were for issues only

casrielasriel:
implementation seems fine, part_type_colour_mix is not implemented but aside from that it is good
image

cobaltgit (commented):
CI needs fixing, other than that looks good to me

PR #383, opened 2026-08-06
Fix normal (non-swap) dup path

MrPowerGamerBR:
While developing Starfait I've noticed that the Butterscotch's `handleDup` path is completely borked

This is a WIP because the swap path also needs to be updated, I haven't done the research on how the swap path should ACTUALLY work

Explanation:

image
image
image
image
image

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -1894,39 +1894,15 @@ static void handleDup(VMContext* ctx, uint32_t instr) {
     }
 #endif
 
-    // Normal dup mode: total bytes to duplicate = (operand + 1) * sizeof(type1)
-    // WAD17+ uses the low 15 bits of operand.
-    // WAD16 and below uses the low 15 bits of operand.
-    // Bit 15 is the swap-mode flag, handled above.
-    int32_t operandCount;
-#if IS_WAD17_OR_HIGHER_ENABLED
-    operandCount = IS_WAD17_OR_HIGHER(ctx) ? (int32_t)(operand & 0x7FFF) : (int32_t)(operand & 0xFF);
-#else
-    operandCount = (int32_t)(operand & 0xFF);
-#endif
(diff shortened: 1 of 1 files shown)


Un1q32:
light mode discord jumpscare

src/vm.c
```diff
@@ -1846,13 +1849,15 @@ static int32_t bytesToSlotCount(VMContext* ctx, int32_t nativeBytes, int32_t sta
     require(remaining == 0); // Byte count must align exactly to slot boundaries
```
Fancy2209, line 1849: The Chapter 4 Test Hits this now

Ananim353:
type1 is the unit of measurement, not the type of each element.

`obj_parallax_castle_top`, `layers[i].xoffset -= n`:

    Push.v self.layers [Array]      ; VARIABLE, 16 bytes
    PushI.e -9                      ; STACKTOP marker, 4 bytes
    Dup.i 5                         ; operand=4 -> 20 bytes
    Push.v self.xoffset [StackTop]  ; eats marker + instance = 20 bytes
    Pop.i.v self.xoffset [StackTop] ; eats the second copy

16 + 4 = the 20 bytes requested, and exactly what both consumers eat. No int32 in the window, nothing mistagged. 
Counting elements takes 5 slots — three too deep.

MrPowerGamerBR:
type1 is the type of the element on the top of the stack, the YoYo Runner does not use tagged elements, so it relies on the type1 to figure out how many bytes will be removed from the stack

The issue is that the compiler sometimes emits a INT32 type even though the top of the stack is VARIABLE

Butterscotch tags the original stack type for each element, but in reality it wouldn't need that, because dup only duplicates the same type always, you can just peek the top of the stack for what is the type that is being used and then figure out the byte size for that

For the integer type, it would require to peek the top to see if it is a variable and, if the type1 is a integer, divide the dup size by 4

PR #385, opened 2026-08-09
argument[0][i] and 32bit colour fix (Platformer mode fix)

Ananim353:
Fix 1: argument[0][i] returned undefined (src/vm.c)

 Fix 2: colours that do not fit a signed int32 turned black (rvalue.h, utils.h, vm_builtins.c)

Also fix Color_lerp truncated instead of rounding (0x7F7F7F vs GameMaker's 0x808080 at t = 0.5),

image


cobaltgit:
Looks good to me

PR #386, opened 2026-08-09
Add macOS runner for building

emiyl:
So we can see if macOS fails to build and also to provide users with macOS binaries from the actions if they don't want to compile it themselves c:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -212,7 +212,7 @@ jobs:
             ccache-macos-${{ matrix.label }}-
   
       - name: Configure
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -33,6 +33,78 @@ jobs:
         with:
           name: butterscotch-glfw-linux-x86_64
           path: build/butterscotch
+          
+  build-glfw-macos-arm64:
+    name: Build (GLFW/macOS arm64)
+    runs-on: macos-latest
```
Un1q32: dont use -latest CI runners because if the version tied to it updates the compiler might update and introduce new warnings, and since we use `-Werror` it would break CI without warning

.github/workflows/build.yml
```diff
@@ -33,6 +33,78 @@ jobs:
...
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: butterscotch-glfw-macos-arm64
+          path: build/butterscotch
+
+  build-glfw-macos-x86_64:
+    name: Build (GLFW/macOS x86_64)
+    runs-on: macos-14
```
Un1q32: use `macos-26-intel` and dont pass the `-DCMAKE_OSX_ARCHITECTURES` argument.  You can omit all the sysctl stuff too

cobaltgit:
Looks good to me

PR #387, opened 2026-08-09
chore(ci): add FreeBSD CI

cobaltgit:


cobaltgit:
I intend on switching this to use cross compilation rather than a VM

cobaltgit:
Ok we've got FreeBSD CI for i386 and AMD64 now.

.github/workflows/build.yml
```diff
@@ -180,6 +180,63 @@ jobs:
       - name: Pack cache
         run: cd ~ && tar cJf ccache.tar.xz .cache/ccache workspacecache.tar
 
+  build-sdl2-freebsd:
+    name: Build (SDL2/FreeBSD ${{ matrix.arch }})
+    runs-on: ubuntu-latest
```
Un1q32: dont use -latest CI runners because if the version tied to it updates the compiler might update and introduce new warnings, and since we use `-Werror` it would break CI without warning

.github/workflows/build.yml
```diff
@@ -180,6 +180,63 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [amd64, i386]
+    env:
+      FREEBSD_SYSROOT: /opt/freebsd-sysroot
+      FREEBSD_MAJOR: 14
+      FREEBSD_MINOR: 3
```
Un1q32, line 193: 14.4 is out, any reason we can't use that?

cobaltgit: This is the earliest version that's on the main freebsd repos and not the archive, and compiling on an older version will be forward-compatible anyway

Un1q32 (approved):

PR #394, opened 2026-08-10
Add small GUI to AppKit version

emiyl:
This also adds a few GUI functions to src/desktop/main.c and platformdefs.h. This is controlled with an ENABLE_GUI option in CMakeLists which defaults to OFF.

The GUI allows the user to open a data.win without a CLI, through a menu in the top bar on macOS. It also allows the user to set the rendering API, the data.win loading mode, the game speed/fast-forward speed, and lazy loading. This can be expanded easily in the future using the gui settings struct.


Un1q32:
I don't like this being folded into the cli desktop backend thing, but it would be good as a seperate thing when the backend refactor is ready.

emiyl:
What's the backend refactor?

Un1q32:
https://github.com/ButterscotchRunner/Butterscotch/pull/374

emiyl:
ahh so cli and gui apps would be done separately?

Un1q32:
yes

emiyl:
Cool beans, I'll bring this back later then

PR #396, opened 2026-08-10
Add appkit port to Makefile

emiyl:
Also fixes an issue where glfw3 wouldn't be found properly on macOS

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ INCLUDES += $(INCLUDE)src/desktop
 ifeq ($(DESKTOP_BACKEND),glfw3)
 GLFW3_CFLAGS := $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --cflags glfw3)
 GLFW3_LIBS := $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs glfw3)
-CFLAGS += $(GLFW3_CFLAGS)
+SYSCFLAGS += $(GLFW3_CFLAGS)
 LIBS += $(GLFW3_LIBS)
 DEFINES += $(DEFINE)USE_GLFW3
 ENABLE_GLAD := 1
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -70,13 +70,16 @@ DEFINES += $(DEFINE)ENABLE_WAD17
...
+SRCS += $(wildcard src/desktop/backends/$(DESKTOP_BACKEND).c src/desktop/backends/$(DESKTOP_BACKEND).m)
 ifeq ($(OS),Windows)
 PKG_CONFIG_FLAGS := --static
 endif
 INCLUDES += $(INCLUDE)src/desktop
 ifeq ($(DESKTOP_BACKEND),glfw3)
-GLFW3_LIBS += $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs glfw3)
+GLFW3_CFLAGS := $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --cflags glfw3)
+GLFW3_LIBS := $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --libs glfw3)
+CFLAGS += $(GLFW3_CFLAGS)
```
Un1q32: dont modify CFLAGS directly, thats kept simple and always the same so the user can change it without worrying about breaking the build.

emiyl: how would you do it instead?

Un1q32: add another variable to the compile step and use that, call it SYSCFLAGS or something

Un1q32:
While we're at it do it for glfw2 as well.
SDL has some issues if you use pkg-config's cflags because it will pass like `-I/usr/include/SDL` which is NOT the directory you want, so don't do SDL

emiyl:
i’ll take a look later, just heading out now, but i remember not having issues compiling for glfw2

PR #397, opened 2026-08-10
Add cflags for SDL

emiyl:
This fixes compiling for SDL2 and SDL3 on my Mac. SDL1 doesn't work but I'm assuming that's an issue with my setup or something, I've just made it match though.


Un1q32:
Screenshot 2026-08-10 at 6 14 20 PM

sdl 1.2 is the only one that doesnt include the right thingy ugh.

PR #399, opened 2026-08-10
Add show_error function

emiyl:
Adds support for `show_error(message: String, abort: bool)`.

### Implementation

The function takes two arguments. The first, the error message, is sent to `runner->showErrorDialogue` if it exists. A check is made before it's called, as showErrorDialogue is not implemented for all platforms. If the abort bool is true, it sets `runner->shouldExit` to true.

Regardless of what happens, the function also fprintf's the error to the console log. 

### Windows, macOS and Linux
Error messages must be implemented per platform. I've created a directory `src/desktop/platform/` with files for `windows.c`, `macos.m` and `gtk3.c`.

- **Windows:** We can call `MessageBoxA`.
- **macOS**: We can use an NSAlert for macOS 10.3 and above, and an NSRunAlertPanel for versions prior to 10.3.
- **Linux** : We can use GTK3 to show the error box.

#### macOS 26.5.1, native NSAlert
image

#### KDE Plasma VM, GTK3
image

### Why not SDL2/3 SDL_ShowSimpleMessageBox?

First of all it looks.. really ugly. And @Un1q32 says that some older versions can segfault really easily. Also I think that targeting GTK3, Windows and macOS should cover >99% of desktop users.

### PS Vita
SceMsgDialogParam is used to show an error message.

image

### Android, Web, PS2 and PS3
Android and PS3 have native error dialogues that could be implemented. For all other devices, I'd like to implement a fallback with something like this.

image

While it's not pretty yet, it uses gamemaker builtin functions to draw to the screen, so we can guarantee that it'll be compatible with all platforms that butterscotch targets.


.github/workflows/build.yml
```diff
@@ -218,6 +218,7 @@ jobs:
             -DWERROR=ON \
             -DDESKTOP_BACKEND=glfw3 \
             -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_C_FLAGS="-Wno-error=unused-but-set-global" \
```
Un1q32: why

emiyl: there was some issue with an unused variable in glad

Un1q32:
I think including gtk and qt isn't worth it and we should just use nuklear or some other solution that doesn't require platform specific apis or new big dependencies.

emiyl:
I've removed qt6 now and GTK3 is an optional dependency that only adds it if it detects that it's installed. There's still a fallback to just print to the console.

If you want minimal dependencies then I think my fallback that uses GameMaker functions to draw the error message would be better than nuklear as we can 100% guarantee it'll work on any platform that Butterscotch works on.

Un1q32:
> If you want minimal dependencies then I think my fallback that uses GameMaker functions to draw the error message would be better than nuklear as we can 100% guarantee it'll work on any platform that Butterscotch works on.

Can we just use that everywhere (even on windows and mac), I think it looks better then messageboxes anyway.

emiyl:
I mean I guess but I haven’t figured out how to properly do that yet, currently it’s just the ugly solution of pausing everything and forcing it to the screen. I think the messages boxes feel more native and are what a user would expect.

PR #401, opened 2026-08-11
replace random with well512a

Un1q32:
Replaces all uses of rand() and srand() with a custom implementation of the well512a algorithm.
This is the same algorithm that yoyo uses, though the way we use it might not match theirs so we prolly don't get the exact same results.

### DIFF
--- a/src/android/main.c
+++ b/src/android/main.c
@@ -363,7 +363,7 @@ static bool startRunnerFromPath(const char* dataWinPath, const char* savesPath, 
         audioSystem = (AudioSystem*) NoopAudioSystem_create();
     }
 
-    Runner* runner = Runner_create(dataWin, vm, renderer, (FileSystem*) overlayFs, audioSystem);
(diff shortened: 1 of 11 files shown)


Un1q32:
This should improve randomness quality on windows, since it's rand() only has 15 bits of randomness and this should have a full 32 bits of randomness.  It also means the randomness quality will be consistent across all platforms.

cobaltgit:
How come you changed from GMLReal to double in a couple cases?

Un1q32:
> How come you changed from GMLReal to double in a couple cases?

~~Accuracy with the percentage stuff, for the random functions that specify a maximum it would be easy to exceed that maximum for some random outputs.~~
Actually I just realized it's fine and doesn't matter, I'll change it back soon.

cobaltgit (approved):

cobaltgit:
Ok regression: aqua and her knife chain attack has a tendency to just disappear at times in ch5 in battle

PR #405, opened 2026-08-12
Suppress deprecation warning in macOS build

emiyl:
Fixes build fail in #386


Un1q32:
Do we not universally disable the macOS deprecation warns?  We should do that instead, Apple loves to deprecate everything and then never remove it.  The only thing deprecation means is the next time there's a new Darwin subsystem (iOS, watchOS, whateverthefuckOS) it won't be in there.

emiyl:
I was thinking it was strange I wasn't getting this warning before lol, I'll take a look at what changed actually

Fancy2209:
Aren't CGamepad and CGamepadExtended both 10.9+?

PR #408, opened 2026-08-12
ARM64, LLVM and Matrices

emiyl:
- Switches Windows runners to use llvm-mingw
- Splits build.yml into four modules
- Adds ARM64, x86_64 and i686 Windows runner matrix
- Adds ARM64, x86_64 and i686 Linux runner matrix
- Fixes a minor compiler issue in SDL2


.github/workflows/build.yml
```diff
@@ -53,41 +67,81 @@ jobs:
...
+
+          make -C /tmp/bzip2-1.0.8 libbz2.a \
+            CC=${{ matrix.triplet }}-gcc \
+            AR=${{ matrix.triplet }}-ar \
+            RANLIB=${{ matrix.triplet }}-ranlib \
+            -j$(nproc)
+
+          sudo cp /tmp/bzip2-1.0.8/libbz2.a /usr/${{ matrix.triplet }}/lib/
+          sudo cp /tmp/bzip2-1.0.8/bzlib.h /usr/${{ matrix.triplet }}/include/
+
```
cobaltgit: We can completely get rid of this step as we statically vendor bzip2 anyway

.github/workflows/build.yml
```diff
@@ -7,92 +7,107 @@ on:
...
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-            -DCMAKE_INSTALL_PREFIX=/usr/x86_64-w64-mingw32 \
-            -DGLFW_BUILD_EXAMPLES=OFF -DGLFW_BUILD_TESTS=OFF -DGLFW_BUILD_DOCS=OFF
-          cmake --build /tmp/glfw/build
-          sudo cmake --install /tmp/glfw/build
-
-      - name: Configure
-        run: cmake -B build -G Ninja -DCMAKE_TOOLCHAIN_FILE=cmake/mingw-w64.cmake -DWERROR=ON -DDESKTOP_BACKEND=glfw3 -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+            "-DBUTTERSCOTCH_COMMIT_HASH=${{ steps.commit-info.outputs.hash }}" \
+            "-DBUTTERSCOTCH_COMMIT_DATE=${{ steps.commit-info.outputs.date }}"
```
cobaltgit: These I imagine go unused too

Un1q32:
Why use llvm?  Is gcc not fine?

emiyl:
> Why use llvm? Is gcc not fine?

For cross compilation it seems easier

Un1q32:
I guess since distros don't package arm64 mingw.

.github/workflows/build.yml
```diff
@@ -7,43 +7,211 @@ on:
...
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: x86_64
+            runner: ubuntu-24.04
+            cc: gcc
+            cxx: g++
+            cflags: ""
```
Un1q32: you can just remove these lines they dont do anything

.github/workflows/build.yml
```diff
@@ -7,43 +7,211 @@ on:
...
+            runner: ubuntu-24.04
+            cc: gcc
+            cxx: g++
+            cflags: "-m32"
+            cxxflags: "-m32"
+            ldflags: "-m32"
+
+          - arch: arm64
+            runner: ubuntu-24.04-arm64
+            cc: gcc
```
Un1q32: we're always using gcc here so just hardcode it

.github/workflows/build.yml
```diff
@@ -7,43 +7,211 @@ on:
...
+          sudo apt-get install -y \
+            cmake \
+            ninja-build \
+            ccache \
+            libglfw3-dev \
+            libbz2-dev \
+            gcc-multilib \
+            g++-multilib
+
+      - name: Build 32-bit Wayland/xkbcommon/GLFW for i686
```
Un1q32: what da hell bruh, is there no 32 bit glfw3 package in repos?  if not just use sdl2 cuz theres no way theres no sdl2 32 bit package

emiyl: ehh it works now tho (and i spent ages trying to get it to compile pls)

Un1q32: I like SDL 2 better anyway since it handles the software renderer better

cobaltgit:
Looks good to me so far

src/desktop/backends/sdl2.c
```diff
@@ -130,9 +130,15 @@ bool platformGetScaledWindowSize(int32_t* outW, int32_t* outH) {
 }
 
 static float platformGetWindowScale(void) {
-    int32_t draw_w, draw_h;
-    int logical_w, logical_h;
-    platformGetWindowSize(&draw_w, &draw_h);
+    int32_t draw_w = 0;
```
Un1q32: we dont need to init these, just add the fail check on its own

CMakeLists.txt
```diff
@@ -267,6 +267,7 @@ if(PLATFORM STREQUAL "cli")
     # GLAD
     add_library(glad STATIC vendor/glad/src/glad.c)
     target_include_directories(glad PUBLIC vendor/glad/include)
+    target_compile_options(glad PRIVATE -Wno-unused-but-set-variable)
```
Un1q32: why

cobaltgit: Idk what that was doing there

Un1q32:
Why have a build-other.yml when we can separate those too?

Un1q32:
build-ps2, build-ps3, build-vita, build-old-compilers, build-c++

PR #417, opened 2026-08-16
Make the vita port use the new backend system

Un1q32:
Removes a lot of duplicated code in the vita sources that we don't need.
Also theoretically makes the vita port compatible with backends other than SDL 2, so you could build it with SDL 3 or GLFW if you wanted.
Should also make the vita compatible with the software renderer.


leap0x7b:
i'm assuming other platforms (ps2, ps3, etc) would get migrated to the new backend system in the future right

Un1q32:
> i'm assuming other platforms (ps2, ps3, etc) would get migrated to the new backend system in the future right

Yes, eventually.  And all new backends should use the new system.

PR #419, opened 2026-08-17
change string_starts_with to use memcmp

basiccube:
#392 changed string_starts_with to use strcmp instead of memcmp to fix a crash when comparing a substring that's longer than the specified string
the problem is, with strcmp, it compares both strings in their entirety rather than just the length of the second string, so something like `string_starts_with("Hello World!", "Hello");` wouldn't return true

this basically reverts it back to using memcmp while adding an additional check for the substring being longer than the string


emiyl:
oh whoops 😅

emiyl (approved):

PR #427, opened 2026-08-18
chore(vita): restore optimisation flags for non-debug builds

cobaltgit:


CMakeLists.txt
```diff
@@ -726,6 +726,7 @@ elseif(PLATFORM STREQUAL "ps3")
     target_link_libraries(mojoal PUBLIC ${SDL2_LIBRARIES})
     target_link_libraries(butterscotch PRIVATE ${SDL2_LIBRARIES} m gcm_sys rsx sysutil io audio rt lv2 io audio bzip2 stb_ds sha1 ps3gl mojoal stb_vorbis)
 elseif(PLATFORM STREQUAL "vita")
+    add_compile_options(-marm)
```
cobaltgit: vitaSDK's gcc was compiled with `--with-mode=thumb`

cobaltgit:
vitaGL builds are now cached

Un1q32:
The optimization flags aren't needed, cmake will add -O3 automatically, and the other stuff is always implied by the vita compiler

Un1q32:
Actually idk about -fomit-frame-pointer maybe keep that in.

cobaltgit:
I'll redo this when the ci rework gets merged

PR #428, opened 2026-08-19
feat(runner): add fps_real builtin and fix fps builtin

cobaltgit:
fixes the camera surface in Alphys's lab freezing in Undertale

image


cobaltgit:
Should we leave deltaTime as a double or would it be safe to have it as GMLReal (for float-real platforms)

Un1q32:
> Should we leave deltaTime as a double or would it be safe to have it as GMLReal (for float-real platforms)

I wonder if it would be better as a float actually, save a little memory and compute time.  We should prolly look thru the codebase for stuff that is double or GMLReal that can be changed to a float without any real chance of fucking things up.  Saves a little bit of memory and maybe compute time for free.

Un1q32:
Actually I looked at the usage and no you cannot change to GMLReal

Un1q32:
float isn't accurate enough for good timekeeping

Un1q32:
dawg i got no idea if this is good or not idk anything about the gml interpreter

Un1q32:
beast mode

PR #359, opened 2026-07-25
feat(vertex): Add vertex and primitives functions

emiyl:
Functions based on Ghidra decomp of GameMaker Runner Windows-x64-VM-Green-2024.14.4.268.

We have triangles!
image

Knight's first attack pattern now appears
image

Slight issue with cropping on the textures? Not too sure but not the end of the world I guess 
image

PR #429, opened 2026-08-19
Software renderer implementation

iProgramMC:
Continuation of #128

Still a draft because I'm not completely happy with the implementation.  However, UNDERTALE mostly works, and I've even been able to play through some of DELTARUNE as well, including the final boss fight of Chapter 3.

(Note: The "stars" attack of the Knight in Deltarune Ch.3 is only possible because I don't implement masking properly.  It's actually not playable in Butterscotch at all because a graphics function isn't implemented)


emiyl:
> Note: The "stars" attack of the Knight in Deltarune Ch.3 is only possible because I don't implement masking properly. It's actually not playable in Butterscotch at all because a graphics function isn't implemented

#359 should fix this 🙏

unknownindividualpersonnel:
History is going to be made when this is merged.

cobaltgit:
> History is going to be made when this is merged.

DEC Alpha NT

unknownindividualpersonnel:
> DEC Alpha NT

who gave you a pile of weed bruh 😭

iProgramMC:
> History is going to be made when this is merged.

Yes indeed

Un1q32:
IMO this is good enough to merge right away, even if not all features are implemented.  I believe undertale is perfect, and deltarune only has a few minor issues.

Un1q32:
Merging now means that new vtable functions could be ported to the software renderer as they're designed in PRs for example, instead of after the fact.  iProgram tends to go on long hiatuses in between working on this, and merging means changes to the renderer don't accumulate tech dept as much, since the person writing the pr is responsible for making their change work on the software renderer, rather than iProgram or one of the maintainers later on after it's merged.

PR #431, opened 2026-08-19
Don't repeat libraries for PS3 build

emiyl:
Libraries were repeated in the PS3 build script, and running them on my machine would give errors. This version only has the libraries once, in the CMakeLists.txt, and seems to work lovely.


emiyl:
The diff shows up weird but it should merge fine

PR #433, opened 2026-08-19
feat: Switch port

cobaltgit:
image

Eden shenanigans


cobaltgit:
Where the fuck is my audio

cobaltgit:
NVM i'm stupid

cobaltgit:
Issues:
- it's upside down and trying to render in widescreen for 4:3 games
- Deltarune STILL doesn't have sdl2 input fuck knows why

src/switch/main.c
```diff
@@ -0,0 +1,49 @@
...
+    args.exitAtFrame = -1;
+#ifdef ENABLE_VM_TRACING
+    args.traceBytecodeAfterFrame = 0;
+#endif
+    args.speedMultiplier = 1.0;
+    args.fastForwardSpeed = 0.0;
+    args.osType = OS_SWITCH;
+    args.profilerFramesBetween = 0;
+    args.loadType = DATAWINLOADTYPE_LOAD_IN_MEMORY_AHEAD_OF_TIME;
+    args.lazyRooms = false;
```
Un1q32: Doesn't this slow down room transitions?  The switch has like 8 gig of ram right, that's enough that we shouldn't need lazy rooms enabled.

Un1q32: Oh wait I'm stupid it says false, just remove the line it's false by default

CMakeLists.txt
```diff
@@ -148,7 +162,7 @@ if(AUDIO_BACKEND STREQUAL "openal")
     add_compile_definitions(USE_OPENAL)
     if(APPLE)
         set(AUDIO_LIBRARIES "-framework OpenAL")
-    elseif(PLATFORM STREQUAL "vita")
+    elseif(PLATFORM STREQUAL "vita" OR PLATFORM STREQUAL "switch")
```
Un1q32, line 180: Switch shouldn't need pthread to use OpenAL, I might be wrong though.

CMakeLists.txt
```diff
@@ -383,6 +409,12 @@ if(PLATFORM STREQUAL "cli" OR PLATFORM STREQUAL "vita")
         )
     else()
         target_link_libraries(butterscotch PRIVATE ${BACKEND_LIBRARIES} ${AUDIO_LIBRARIES} ${PLATFORM_LIBRARIES})
+
+        if (PLATFORM STREQUAL "switch")
+            target_link_libraries(butterscotch PRIVATE GLESv2 stdc++)
```
Un1q32, line 424: I don't think you can call this twice, I think you need to put switch in an else if before the final else

cobaltgit: Seems like you can, it just appends the libraries right?

Un1q32: Last time I tried it errored

src/switch/main.c
```diff
@@ -14,8 +14,8 @@
 
 int main(int argc, char* argv[]) {
     (void)argc;
-    setbuf(stderr, NULL);
-    setbuf(stdout, NULL);
+    setvbuf(stdout, NULL, _IONBF, 0);
```
Un1q32: why? They should be equivalent.

CMakeLists.txt
```diff
@@ -96,6 +106,10 @@ option(ENABLE_WAD16 "Enable support for WAD Version 15/16" ON)
 option(ENABLE_WAD17 "Enable support for WAD Version 17" ON)
 option(ENABLE_LEGACY_GL "Enable the legacy OpenGL renderer" ON)
 option(ENABLE_MODERN_GL "Enable the modern OpenGL renderer" ON)
+if(PLATFORM STREQUAL "switch")
+    # The Switch only supports the modern GL renderer (GLES3 via SDL2)
```
Un1q32: I'm like 80% sure that the switch can do desktop OpenGL, so legacy-gl should work.

cobaltgit: It could probs work with glad, I'll have a look

src/switch/main.c
```diff
@@ -29,10 +29,6 @@ int main(int argc, char* argv[]) {
     args.fastForwardSpeed = 0.0;
     args.osType = OS_WINDOWS;
     args.profilerFramesBetween = 0;
-    args.loadType = DATAWINLOADTYPE_LOAD_IN_MEMORY_AHEAD_OF_TIME;
```
Un1q32: This should be kept

CMakeLists.txt
Un1q32: I don't think so, switch doesn't have shared libraries and definitely no shared libGL, so glad won't work.

cobaltgit:
Eden segfaults with OS_WINDOWS but not OS_SWITCH, Ryujinx doesn't boot it at all

Un1q32:
do you have real hardware to test with?

cobaltgit:
Nah, I wish I did tho

For legal reasons I am totally lying

CMakeLists.txt
```diff
@@ -229,6 +239,18 @@ if(PLATFORM STREQUAL "cli" OR PLATFORM STREQUAL "vita")
...
+        set(VM_GML_PROFILER_DEFAULT OFF)
+        set(VM_TRACING_DEFAULT OFF)
+        set(VM_OPCODE_PROFILER_DEFAULT OFF)
+        set(VM_STUB_LOGS_DEFAULT OFF)
+    elseif(PLATFORM STREQUAL "switch")
+        if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
+            add_compile_options(-ffast-math)
+        endif()
+
+        add_compile_definitions(__SWITCH__)
```
Un1q32: Doesn't the dkp compiler define this by default?

Un1q32: test if removing this line breaks the build

Un1q32:
Is this done?

cobaltgit:
Not quite, unless I merge it in a half-done state and rip my hair out over inputs not working and forced widescreen rendering

Un1q32:
What about SDL 3?

Un1q32:
or glfw3, the switch has a glfw3 port in dkp repos

cobaltgit:
image

All is good in the world

PR #439, opened 2026-08-21
chore(ci): LTO for all!

cobaltgit:
Our use of ccache should offset LTO build-time increase sooo..here we are
I'll push again when everything succeeds just to test build time with a warm cache

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -138,6 +138,7 @@ jobs:
             -DWERROR=ON
             -DBACKEND=${{ matrix.desktop_backend }}
             -DCMAKE_BUILD_TYPE=Release
+            -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
             -DCMAKE_C_COMPILER_LAUNCHER=ccache
             -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
           )
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -141,7 +141,6 @@ jobs:
               VERBOSE=1 \
               CC='ccache i486-w64-mingw32-gcc' \
               CFLAGS='-O3 -flto -DNDEBUG -fomit-frame-pointer' \
-              LDFLAGS='-flto' \
               BACKEND=glfw2 \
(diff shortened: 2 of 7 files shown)


cobaltgit:
PS2 artifacts seem to be a bit inflated as a result of LTO, not that it matters so long as performance isn't impacted negatively

.github/workflows/build-windows.yml
```diff
@@ -137,7 +140,8 @@ jobs:
               NO_COLOR=1 \
               VERBOSE=1 \
               CC='ccache i486-w64-mingw32-gcc' \
-              CFLAGS='-O3 -DNDEBUG -fomit-frame-pointer' \
+              CFLAGS='-O3 -flto -DNDEBUG -fomit-frame-pointer' \
+              LDFLAGS='-flto' \
```
Un1q32: not needed

cobaltgit: As in the LDFLAGS? I thought you also needed to have them set to flto

Un1q32: no you dont, the objects created will be LTO objects and the linker will figure it out.

cobaltgit:
Should be ready to go! Reduced binary size for some desktop builds, PSVita, PS3 etc

.github/workflows/build-linux.yml
```diff
@@ -180,6 +180,9 @@ jobs:
             -DCMAKE_C_COMPILER=loongarch64-linux-gnu-gcc \
             -DCMAKE_CXX_COMPILER=loongarch64-linux-gnu-g++ \
             -DCMAKE_INSTALL_PREFIX="${{ github.workspace }}/vendor/SDL2/install" \
+            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_POLICY_DEFAULT_CMP0069=NEW \
```
Un1q32, line 184: ??

cobaltgit: Oh cmake complains about that for some reason with it being an older cmake file, just enforcing the new way of adding LTO to cflags

Fancy2209:
LTO Breaks PS3 don't use it there kthx

cobaltgit:
Moving to a different pr

PR #441, opened 2026-08-21
chore(ci): add riscv64 to FreeBSD CI

cobaltgit:


.github/workflows/build-freebsd.yml
```diff
@@ -46,7 +51,29 @@ jobs:
...
+          path: vendor/SDL2
+
+      - name: Build SDL2 from source (RISC-V)
+        if: matrix.arch == 'riscv'
+        run: |
+          cmake -S vendor/SDL2 -B vendor/SDL2/build -G Ninja \
+            -DCMAKE_TOOLCHAIN_FILE=cmake/freebsd-llvm.cmake \
+            -DCMAKE_C_COMPILER_TARGET=${{ matrix.target_triple }} \
+            -DCMAKE_CXX_COMPILER_TARGET=${{ matrix.target_triple }} \
+            -DSDL_SHARED=ON \
```
Un1q32: shouldn't it be static?  If FreeBSD doesn't have an SDL 2 package we can download then it would be better to static link so riscv users don't need to do their own build from source.

PR #448, opened 2026-08-23
chore(ci): i386 and PowerPC Mac OS crossbuilds

cobaltgit:


unknownindividualpersonnel:
coolest fucking pr in the history of butterscotch right here folks

cobaltgit:
There'll be a better way to build these eventually (using gcc 16 for both)

PR #452, opened 2026-08-26
links and links and links

cobaltgit:


README.md
```diff
@@ -87,6 +83,45 @@ The following compilers have been tested to successfully build butterscotch, old
 * TinyCC 0.9.27 and up
 * MSVC 4.0 and up
 
+## Download Butterscotch
+
+Below are direct links to download the latest auto-build of Butterscotch for your platform.
+
+### Desktop
+
+| OS | x86_64 | i486 | arm64 | armhf | PowerPC | PowerPC64 | PowerPC64 little-endian | RISC-V | LoongArch64 | MIPS | MIPS little-endian | MIPS64 | MIPS64 little-endian |
```
Un1q32: we should call i486 here i386, since macOS will be called i386

Un1q32: only the label on top tho, keep the others as i486

cobaltgit: Wouldn't macOS be i686 cause Yonah/Pentium M? Idk

Un1q32: technically yeah but everyone says i386 because apple says i386

cobaltgit: True

PR #460, opened 2026-08-29
perf(gl): actual CPU state tracking wrappers

cobaltgit:
Should benefit the OpenGL 2.x/legacy-gl paths more than 3.0+ but here we are. Less pipeline stalls yippee!

### DIFF
--- a/src/gl_common/gl_state.h
+++ b/src/gl_common/gl_state.h
@@ -1,16 +1,18 @@
 #ifndef _BS_GL_STATE_H
 #define _BS_GL_STATE_H
 
+#define GL_MAX_TEXTURE_2D 32
+
 static GLenum activeTextureSlot = GL_TEXTURE0;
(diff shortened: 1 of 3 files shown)


src/gl_common/gl_state.h
```diff
@@ -0,0 +1,193 @@
...
+            currentDrawFb = framebuffer;
+            break;
+        default:
+            return;
+    }
+
+    glBindFramebuffer(type, framebuffer);
+}
+#undef glBindFramebuffer
+#define glBindFramebuffer cached_glBindFramebuffer
```
Un1q32, line 109: this will be incompatible with the gl wrappers stuff

cobaltgit: Ah I included it after gl_wrappers so it should wrap the framebuffer compatibility wrapper right?

PR #493, opened 2026-09-03
refactor(runner): make particles use the global RNG

cobaltgit:
Same seed as global PRNG. Previous impl used Xorshift32 (faster but MUCH more deterministic) and was a static var in vm_builtins rather than correctly being an attribute of the runner struct

TODO: is there really a separate PRNG stream for particles?

### DIFF
--- a/src/runner.c
+++ b/src/runner.c
@@ -2371,7 +2371,9 @@ Runner* Runner_create(DataWin* dataWin, VMContext* vm, Renderer* renderer, FileS
     renderer->runner = runner;
     runner->viewportW = 1;
     runner->viewportH = 1;
+    runner->initSeed = randomSeed;
     runner->random = Random_create(randomSeed);
+    runner->particleRngState = Random_create(randomSeed);
 
     repeat(MAX_SURFACES, i) {
         runner->surfaceStack[i] = -1;
--- a/src/runner.c
+++ b/src/runner.c
@@ -2372,7 +2372,6 @@ Runner* Runner_create(DataWin* dataWin, VMContext* vm, Renderer* renderer, FileS
     runner->viewportW = 1;
     runner->viewportH = 1;
     runner->random = Random_create(randomSeed);
-    runner->particleRngState = Random_create(randomSeed);
 
     repeat(MAX_SURFACES, i) {
         runner->surfaceStack[i] = -1;
--- a/src/runner.c
+++ b/src/runner.c
@@ -2371,7 +2371,6 @@ Runner* Runner_create(DataWin* dataWin, VMContext* vm, Renderer* renderer, FileS
     renderer->runner = runner;
     runner->viewportW = 1;
     runner->viewportH = 1;
-    runner->initSeed = randomSeed;
     runner->random = Random_create(randomSeed);
     runner->particleRngState = Random_create(randomSeed);
 
--- a/src/runner.h
+++ b/src/runner.h
@@ -782,11 +782,6 @@ struct Runner {
     uint64_t gameStartTime;
 
     Random random;
-
-    // Particles draw from their own random stream instead of rand(). Sharing rand() would make every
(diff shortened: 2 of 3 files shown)


src/runner.c
```diff
@@ -2371,7 +2371,9 @@ Runner* Runner_create(DataWin* dataWin, VMContext* vm, Renderer* renderer, FileS
     renderer->runner = runner;
     runner->viewportW = 1;
     runner->viewportH = 1;
+    runner->initSeed = randomSeed;
```
Un1q32: just use 0 for particles and dont store the initseed

Un1q32: actually nevermind, keep it as is i changed my mind

src/runner.h
```diff
@@ -781,7 +781,16 @@ struct Runner {
...
 
+    // Initial random seed for global and particle RNGs.
+    uint32_t initSeed;
+
     Random random;
+
+    // Particles draw from their own random stream instead of rand(). Sharing rand() would make every
+    // particle spawn shift the sequence the game itself sees, so merely adding a particle effect to a
+    // scene would change unrelated randomised behaviour (and every seeded screenshot test with it).
+    // The trade-off is that --seed and randomize() do not reach particles.
```
Un1q32:
```suggestion
```

src/vm_builtins.c
```diff
@@ -17555,25 +17557,13 @@ static RValue builtin_sprite_get_info(VMContext* ctx, RValue* args, int32_t argC
...
-static uint32_t g_particleRngState = PARTICLE_RNG_SEED;
-
-static uint32_t particleRandomBits(void) {
-    uint32_t x = g_particleRngState;
-    x ^= x << 13;
-    x ^= x >> 17;
-    x ^= x << 5;
-    g_particleRngState = x;
-    return x;
+static uint32_t particleRandomBits(Runner* runner) {
```
Un1q32: just remove this function and inline it everywhere

src/vm_builtins.c
```diff
@@ -3229,6 +3229,7 @@ static RValue builtin_random_set_seed(MAYBE_UNUSED VMContext* ctx, RValue* args,
     GMLReal seed = RValue_toReal(args[0]);
     bool fixRangeBug = RValue_toBool(args[1]); 
     Random_setSeed(&ctx->runner->random, (uint32_t) seed);
+    Random_setSeed(&ctx->runner->particleRngState, (uint32_t) seed);
```
Un1q32: does the yoyo runner do this?

cobaltgit: Tryna think whether it has a full separate RNG state for particles let alone this before i merge, hmm

src/vm_builtins.c
```diff
@@ -18070,7 +18055,7 @@ void Particles_freeAll(Runner* runner) {
     runner->particleTypePool = nullptr;
     // Reached on game_restart as well as shutdown, so put the stream back to where it started:
     // a restarted game should produce the same particles as the first run.
-    g_particleRngState = PARTICLE_RNG_SEED;
+    Random_setSeed(&runner->particleRngState, 0);
```
Un1q32: I think you can just do nothing here, this function is only called when the runner is about to be destroyed right?  So another Runner_create will always come before new particles happen, therefore resetting the seed is redundant.

cobaltgit: game_restart() doesn't tear down the runner i don't think

cobaltgit: Well it calls Runner_reset which doesn't reset our RNG states (i think the official runner does tho idk)

Un1q32: In that case maybe we should store the seed and reset it properly like you were doing before.

cobaltgit: Oh apparently it doesn't reseed the rng

Un1q32 (commented):
Lgtm, test and update the test screenshots before merging.

PR #536, opened 2026-09-14
refactor(ci): only run x86_64 and i486 builds for pull requests

cobaltgit:


.github/workflows/build-windows.yml
```diff
@@ -5,6 +5,7 @@ on:
 
 jobs:
   build-arm-windows:
+    if: inputs.full-matrix
```
Un1q32: keep arm windows, serves as an llvm-mingw test

PR #540, opened 2026-09-15
Android: option to handle physical controllers sideways

GlitchedCode:
For games like shmups that benefit more from portrait orientation than landscape (e.g. ZeroRanger), it could be desirable to end users to have the possibility to handle telescopic controllers sideways (especially in cases such as the controller not fitting in a different orientation, or when using the USB-C port). This PR adds the `physicalControllerSideways` option to game setup and metadata related modules, and offers a basic implementation of this functionality through `GamepadRouter`. If there's a better way to approach this last part any feedback is welcome.


GlitchedCode:
currently has issues on my zeroranger fork. i suspect due to multiple inputs being simultaneously sent to the underlying VM

PR #548, opened 2026-09-19
FFmpeg Video implementation

casrielasriel:
Implements video functions and playback using FFmpeg.

This is basically just https://github.com/ButterscotchRunner/Butterscotch/tree/video, but instead of kitchensink it uses FFmpeg. 

I have tested Modern and Legacy GL, Switch, and Vita. Video playback and audio works well on all of them.

The following functions are implemented:
- video_open
- video_close
- video_draw
- video_pause
- video_resume
- video_get_format
- video_get_status
- video_get_duration
- video_get_position
- video_enable_loop
- video_set_volume

I tested on both DELTARUNE Chapter 3 (feacher) and DELTARUNE Chapter 5 (flowery) and it plays the video and audio fine.


casrielasriel:
Tested switch, currently does not work. Will fix and the fix should fix vita as well

casrielasriel:
Once i fix that though (tomorrow im tired) this should be good to merge, just gotta fix that and clean this up a bit

casrielasriel:
okay that was an easy fix. if this works on vita then all i need to do is clean up some stuff and yheah

casrielasriel:
[Screencast_20260919_070012.webm](https://github.com/user-attachments/assets/56b36dcd-5136-4c8b-a795-04d352ba632f)

Feacher on switch

casrielasriel:
Vita works! this can be merged now

Datboi6-2:
Youtuber

casrielasriel:
Get the hell out bro:sob:

Un1q32:
Lemme test first

.github/workflows/build-console.yml
```diff
@@ -181,6 +181,43 @@ jobs:
...
+            --disable-programs --disable-doc --disable-network \
+            --disable-everything \
+            --enable-decoder=aac,h264 \
+            --enable-demuxer=mov \
+            --enable-parser=aac,h264 \
+            --enable-protocol=file \
+            --enable-swscale --enable-swresample \
+            --enable-pthreads \
+            --arch=armv7-a --cpu=cortex-a9 --disable-armv6t2 --target-os=none \
+            --extra-cflags=" -std=gnu11 -Wno-error=implicit-function-declaration -Wno-error=int-conversion -Wno-error=incompatible-pointer-types -Wl,-q -O2 -ftree-vectorize -fomit-frame-pointer -ffast-math -D_BSD_SOURCE" \
```
Un1q32: Don't use -ffast-math in ffmpeg, we don't know if its compatible.  it might cause subtle issues that are hard to diagnose.
Also, why not use O3 instead of O2?  Does it increase binary size a lot?

Un1q32:
If ffmpeg is disabled at build time and the game calls video_open, what happens?  Does it softlock like before or does it skip the video and continue as if it had played?

Makefile
```diff
@@ -134,6 +134,15 @@ DISABLE_MODERN_GL := 1
 DEFINES += $(DEFINE)USE_NOOP
 endif
 
+# FFmpeg (video playback)
+FFMPEG_CFLAGS := $(shell $(PKG_CONFIG) $(PKG_CONFIG_FLAGS) --cflags libavformat libavcodec libavutil libswscale libswresample 2>/dev/null)
```
Un1q32, line 140: What do we need all these for?  I thought we only need libavcodec.

src/vm_builtins.c
```diff
@@ -11710,6 +11711,672 @@ static RValue builtin_merge_color(MAYBE_UNUSED VMContext* ctx, RValue* args, MAY
...
+
+    d->eof = false;
+    d->finished = false;
+    d->paused = false;
+    d->lastPts = AV_NOPTS_VALUE;
+    return true;
+}
+
+static VideoDecoderVtable ffmpegVideoDecoderVtable = {
+    .init = ffmpegVideoDecoderInit,
```
Un1q32: This is incompatible with gcc 2.95, I don't know why CI passes

src/vm_builtins.c
```diff
@@ -11710,6 +11711,672 @@ static RValue builtin_merge_color(MAYBE_UNUSED VMContext* ctx, RValue* args, MAY
     return RValue_makeReal((GMLReal) Color_lerp(col1, col2, amount));
 }
 
+#if defined(BUTTERSCOTCH_FFMPEG) && (defined(ENABLE_MODERN_GL) || defined(ENABLE_LEGACY_GL))
```
Un1q32: Why is it guarded behind `(defined(ENABLE_MODERN_GL) || defined(ENABLE_LEGACY_GL))`?  Is it not renderer independent?  I looked thru the code a bit and didn't see any OpenGL calls.

.github/workflows/build-console.yml
casrielasriel: -o3 does bloat code
and yeah, i will remove --ffast-math let me do that real quick

cobaltgit: Tbf it's a tradeoff: code size or performance

Makefile
casrielasriel: libvaformat opens the file and detects the format,
libvacodec decodes H.264 frames into raw YUV
libswscale converts YUV to RGBA for gpu upload
libswresample resamples audio
libautil is core math and other stuff

.github/workflows/build-console.yml
casrielasriel: yeah, i would say stick with -O2, this is generally fine

casrielasriel:
> If ffmpeg is disabled at build time and the game calls video_open, what happens? Does it softlock like before or does it skip the video and continue as if it had played?

it should return undefined, for me it gives the dialogue that the video doesn't work on a fresh save

.github/workflows/build-console.yml
Un1q32: if the vita is good enough to decode at full speed then its prolly fine, tho NEON prolly helps a lot with that.  Pentium 3 might fare worse.

src/vm_builtins.c
```diff
@@ -11710,6 +11711,672 @@ static RValue builtin_merge_color(MAYBE_UNUSED VMContext* ctx, RValue* args, MAY
...
+    return gotFrame ? 0 : 1;
+}
+
+static void ffmpegVideoDecoderDraw(VideoDecoder* decoder, Runner* runner, int32_t surfaceId) {
+    FfmpegVideoDecoder* d = (FfmpegVideoDecoder*)decoder->impl;
+    if (d == nullptr || d->drawPixels == nullptr || runner->renderer == nullptr) return;
+    GLRenderer* gl = (GLRenderer*)runner->renderer;
+    if (gl->surfaceTexture == nullptr || surfaceId < 0 || (int32_t)gl->surfaceCount <= surfaceId) return;
+    if (gl->surfaceTexture[surfaceId] == 0) return;
+    glBindTexture(GL_TEXTURE_2D, gl->surfaceTexture[surfaceId]);
```
Un1q32: Oh heres the OpenGL.  We should have a renderer vtable function for this so it can work in software and other renderers.

casrielasriel: I mean yeah i can do this

Un1q32:
Instead of the ffmpeg code being inline in vm_builtins.c, it should live in src/video/ffmpeg/ffmpeg.c.

Un1q32:
There should be a VIDEO_BACKEND cmake/make option that can be set to ffmpeg to use ffmpeg, or none to use none.

casrielasriel:
> Instead of the ffmpeg code being inline in vm_builtins.c, it should live in src/video/ffmpeg/ffmpeg.c.

sounds good, will do

casrielasriel:
> There should be a VIDEO_BACKEND cmake/make option that can be set to ffmpeg to use ffmpeg, or none to use none.

will also do. going to mark as draft for now and will umark when done

cobaltgit:
Dude y'know Kris flirts with EVERY enemy, right?

casrielasriel:
image

casrielasriel:
> There should be a VIDEO_BACKEND cmake/make option that can be set to ffmpeg to use ffmpeg, or none to use none.

done

casrielasriel:
[Screencast_20260919_180921.webm](https://github.com/user-attachments/assets/53644ed6-3e7e-4ce4-8b6c-ae49837a332e)

if FFmpeg is off:

src/noop_renderer.c
```diff
@@ -317,6 +318,7 @@ Renderer* NoopRenderer_create(void) {
     noopVtable.surfaceFree = noopSurfaceFree;
     noopVtable.surfaceCopy = noopSurfaceCopy;
     noopVtable.surfaceGetPixels = noopSurfaceGetPixels;
+    noopVtable.surfaceUploadPixels = noopSurfaceUploadPixels;
```
Un1q32: just don't require it to not be null, and leave it unset in the noop renderer.

Un1q32: do a check at the call and if the function pointer is null then assume no-op

Un1q32:
remove video.c and inline that stuff in vm_builtins.c, only the backend dependent stuff should be outside vm_builtins.c

casrielasriel:
> remove video.c and inline that stuff in vm_builtins.c, only the backend dependent stuff should be outside vm_builtins.c

> just don't require it to not be null, and leave it unset in the noop renderer.

> do a check at the call and if the function pointer is null then assume no-op


addressed, thanks for the feedback
implementing 2 extra functions then this should be good

casrielasriel:
boom its ready, if there are any nitpicks with it lmk and i'll fix it

Crazywastaken:
interesting, so it just skips the video automatically for 

> Screencast_20260919_180921.webm
> 
> if FFmpeg is off:
 
 this works by "playing" a very short video and telling the game that its over???? also how would it work for flowery cause the vid is split into an ogg and an mp4 file.

casrielasriel:
> interesting, so it just skips the video automatically for
> 
> > Screencast_20260919_180921.webm
> > if FFmpeg is off:
> 
> this works by "playing" a very short video and telling the game that its over???? also how would it work for flowery cause the vid is split into an ogg and an mp4 file.

it wouldnt be any different

Un1q32:
What ffmpeg versions has this been tested with?  Obviously the latest 9.0 version should work but what about 8.x, 7.x, and 6.x versions?

casrielasriel:
only 6 and 7 (Please dont make a 67 joke.)

Un1q32:
wait you tested on 6 and 7 but not 8 or 9?

casrielasriel:
tested 9 too mb

Un1q32:
ok so 8 prolly works then.

casrielasriel:
would assume so

casrielasriel:
chicken

cobaltgit:
chicken jockey

Un1q32:
lgtm

Un1q32 (approved):

casrielasriel:
yay

unknownindividualpersonnel:
> chicken

we'd like to take a moment to announce that- AAAA-

SplaatSites:
> > chicken
> 
> we'd like to take a moment to announce that- AAAA-

All our food keeps BLOWING U-

unknownindividualpersonnel:
> > > chicken
> > 
> > 
> > we'd like to take a moment to announce that- AAAA-
> 
> All our food keeps BLOWING U-

great meme

